PR #38, opened 2020-10-27
Very WIP seL4test stuff for github-actions

LukeMondy:
Modelled after our Bamboo ConfigAsCode setup, the idea was that we can invoke the python script with certain conditions, e.g., all ARM6 boards, or all MCS platforms, and it would either spit out a script to compile them all (that the action would then invoke), or it would do the compilation itself.
Then the Github Action matrix option could be used to split up compilation of the different types of builds nicely.

I only got part of the way through, not sure if it's the right way to go or not.


LukeMondy:
Ah, looking through the code again, I had gotten up to doing `PlanVarients` - essentially, you setup a base `Plan`, and then generate variations off that base.
Again, I was basing this off the Bamboo ConfigAsCode implementation, so there very well could be a simpler way to do things - but it was a good exercise in understanding what we have in Bamboo!

lsf37:
Closing this for a new approach described in #60

PR #40, opened 2021-01-06
thylint action

lsf37:
A GitHub action to lint Isabelle theories for unwanted outer syntax commands like `sorry` or commands that are interactive-only such as `sledgehammer` and `thm`.

### DIFF
--- a/thylint/thylint.py
+++ b/thylint/thylint.py
@@ -205,7 +205,7 @@ def match_chunk(line, chunk, offset, line_num, matches):
                             'end_column': offset+match.end(),
                             'line_content': line,
                             'title': title,
-                            'message':  msg})
+                            'message': msg})
 
 
 def lint_file(file_name):
--- a/thylint/thylint.py
+++ b/thylint/thylint.py
@@ -293,16 +293,21 @@ def main():
     args = parser.parse_args()
 
     matches = []
+    failures = False
     for file in args.files:
-        if args.allfiles or file.endswith('.thy'):
-            matches += lint_file(file)
+        if args.all_files or file.endswith('.thy'):
+            try:
+                matches += lint_file(file)
+            except IOError:
+                print('IO error for file "{0}"'.format(file), file=sys.stderr)
+                failures = True
 
     if args.json:
         print(matches_to_json(matches))
     else:
         print_matches(matches)
 
-    sys.exit(matches != [])
+    sys.exit(failures or matches != [])
 
 
 if __name__ == '__main__':
--- a/thylint/thylint.py
+++ b/thylint/thylint.py
@@ -160,7 +160,13 @@ style_words = (
     ])
 )
 
-all_words = [diag_words, find_proof_words, sorry_words, axiom_words, style_words]
+warnings = {
+    'diag': diag_words,
+    'find-proofs': find_proof_words,
+    'sorry': sorry_words,
+    'axiom': axiom_words,
+    'style': style_words,
+}
 
 
 def print_matches(matches):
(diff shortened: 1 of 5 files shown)


lsf37:
This is now ready for review. An example PR with output is seL4/l4v#188.

Unfortunately, the annotations in files directly currently only work for PRs that are not made from a fork. The GitHub permission model doesn't allow those PRs to generate annotations. We could write a GitHub App later that can be given independent permissions to enable that, but that's a bit more work. For fork-PRs, the action will still produce warnings in the action output, only the annotation task will fail.

thylint/thylint.py
```diff
@@ -0,0 +1,317 @@
...
+    return re.compile('|'.join(r'\b(?<!\')(?:{0})\b(?!\')'.format(x) for x in words))
+
+
+diag_words = (
+    # title
+    "Interactive diagnostic command",
+    # message
+    "This command is usually used interactively only and should only be checked in for "
+    "demonstration purposes.",
+    # word(ish) regexps that should be flagged as warnings
```
corlewis, line 39: Out of curiosity, do you have a canonical source for this list, or is it 'just' a list of commands that you thought should be flagged.

The follow-up question is whether it's worth other people trying to think of more unwanted commands.

thylint/thylint.py
```diff
@@ -0,0 +1,317 @@
...
+
+sorry_words = (
+    # title
+    "Unfinished proof",
+    # message
+    "This command indicates an unfinished or aborted proof.",
+    # word(ish) regexps that should be flagged as warnings
+    regexp_of([
+        '\\\<proof>',
+        'sorry',
```
corlewis, line 134: I can see `sorry` being flagged becoming annoying in long running branches such as rt. I agree that it makes sense to have it as a warning by default, but it would be very nice if there was a structured way to configure it per branch.

thylint/thylint.py
```diff
@@ -0,0 +1,317 @@
...
+
+axiom_words = (
+    # title
+    "Axioms",
+    # message
+    "Locales or definitions should usually be preferred to axioms, because axioms may make the "
+    "logic inconsistent. Merely introducing a new constant (without new laws) is fine, though.",
+    # word(ish) regexps that should be flagged as warnings
+    regexp_of([
+        'axiomatization',
```
corlewis, line 147: I guess most pull requests don't touch the existing axiomatizations, so it will be rare for one to show up in a diff. If it ever becomes a problem though it could be worth adding syntax for telling the linter to ignore part of a file.

thylint/thylint.py
```diff
@@ -0,0 +1,317 @@
...
+    Updates argument 'matches' by side effect.
+    """
+    for (title, msg, regex) in all_words:
+        for match in regex.finditer(chunk):
+            matches.append({'line': line_num,
+                            'start_column': offset+match.start(),
+                            'end_column': offset+match.end(),
+                            'line_content': line,
+                            'title': title,
+                            'message':  msg})
```
corlewis, line 208:
```suggestion
                            'message': msg})
```

corlewis (approved):
This looks like a great idea to me! I don't think it's an issue that annotations don't work for PRs from a fork. The action output should already be very helpful, and if people really want the annotations it's easy to switch to a branch model for PRs.

I do have to admit to not being very familiar with either the git or python parts of this PR, but you've made it incredibly readable so that wasn't an issue! I especially like the comments, you reached the perfect level of detail.

thylint/thylint.py
lsf37: I went through the output of `print_commands` and took everything that was interactive-only.

thylint/thylint.py
lsf37: Yes, I should probably add a few options to remove specific classes of matches so that we can configure it at the use site.

- [ ] add options

thylint/thylint.py
lsf37: That one is probably harder, although possible. We're always fine to just ignore the output, though, we don't have to make it a requirement to reach zero annotations.

lsf37: I'll leave this one for another iteration.

PR #41, opened 2021-01-11
thylint: only report on lines in diff

lsf37:
This PR adds an option to thylint to only report on lines that are mentioned in the diff of a pull request (in added or changed lines).


corlewis (approved):
Very nice!

PR #43, opened 2021-02-03
style: produce predictable paths

lsf37:
Make sure `find` produces relative, not absolute paths.

Related to seL4/camkes-tool#42 and should fix the issue when the corresponding sel4_tools PR is out.

### DIFF
--- a/style/steps.sh
+++ b/style/steps.sh
@@ -34,7 +34,7 @@ then
     ${SEL4_TOOLS}/misc/style.sh
 else
   # not running in pull request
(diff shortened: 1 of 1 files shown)


lsf37:
Basically, in the `push` case, the `.stylefilter` specs weren't working, because `find` in `style-all.sh` produces unpredictable absolute paths, and the matcher isn't sophisticated enough to figure out the relation to the repo root. This PR makes sure the paths are all relative to the current dir, which is the repo root.

oliver-wm (approved):

PR #49, opened 2021-02-24
Make PR Queue node package

xurtis:
This ensures that the package dependencies are correctly described so that actions provides them in the runtime,


lsf37:
For the license check, it's probably easiest to add a line to the `.reuse/dep5` file for `package.json` etc

lsf37 (approved):
Apart from the license check fixup, lgtm!

xurtis:
How do I add license to the JSON files?

PR #51, opened 2021-03-01
Fixes for PR queue

xurtis:
This fixes a bug and also tries to reduce the number of queries made to GitHub each time this is run.


lsf37 (approved):
👍

lsf37:
I'm putting up a separate PR to ignore all node_modules in the link checker, because it's of course going to keep failing now.

PR #52, opened 2021-03-02
exclude all node-modules from link check

lsf37:
(for this repo, not globally)


lsf37:
Right, the -x option on the checker excludes URLs, not local files to check..w

PR #56, opened 2021-03-28
provide manifest option for repo

lsf37:
Currently specifically for the verification manifest, defaulting to `devel.xml`

This now means that the preprocess test on bamboo and GitHub will be in sync, both using `devel.xml` in `verification-manifest`.


lsf37:
(and I should do something about the link test in general, the exclude mechanism doesn't work as I thought it would)

Xaphiosis (approved):

Issue #16, opened 2020-06-15
Portable shell check is slow


lsf37:
Installing dependencies seems to take forever compared to what the action does.

A docker action would probably be faster.

PR #57, opened 2021-04-10
Faster bashisms

lsf37:
avoid unnecessary dependencies, in particular add the upstream `checkbashisms` script to this repo to avoid downloading a whole lot of unused Debian dev packages.

Run time is now in the order of seconds as opposed to ~1min. (The test that shows up in this PR is the old version of the workflow, the new version becomes active when merged).


lsf37:
This will fix #16

xurtis (approved):

PR #67, opened 2021-06-18
bashisms: update link

lsf37:
Only the link changed, the `checkbashisms` script has remained unchanged between versions.

(This fixes the failed link check)


wom-bat (approved):
Looks good.

Issue #69, opened 2021-06-21
seL4: add C Parser action 


lsf37:
A CI action that runs the C parser on the supported platforms. The preprocess test does some of that already, but for fewer combinations.

See `sel4_test` family `C Parser` in Bamboo.

lsf37:
This already draws in all of the platform/build/run definition fun, so it'll take a while, but I guess it's as good a place as any to start that.

lsf37:
Adding a bit of progress/thinking here:
- most config is captured by a bunch of yaml files
- filled in and combined in a python script
- C parser test itself is fully sequential, so all driven by one python script
- for parallel jobs later, script would get keys from a GitHub-level build matrix, and would translate that into one run or possibly multiple combined runs (e.g. all armv8 combinations or all combinations for one board or something like that)

The C parser action itself will need access to the standalone parser, which we already build a docker image for in the preprocess action.

So, things remaining to do:
 - [ ] merge seL4/seL4-CAmkES-L4v-dockerfile#18 to remove separate `-riscv` images
 - [ ] from preprocess docker image: factor out separate C parser builder that can be applied independently
 - [ ] put this together with python script into a test runner docker file
 - [ ] set up build/deploy job for that in ci-actions

PR #96, opened 2021-06-27
GitHub action for C parser runs

lsf37:
Adds an action for running the verification C parser on seL4 pull requests.

- add seL4 build platform definitions + python module for accessing them
- add build definitions for parser runs + python module for them (a simple first version)
- add the action itself

The action is currently completely sequential and includes both build + the parser run. Should be equivalent to the current Bamboo job for the same.

It might make sense to add MCS variants to this, which could be run in an concurrent job, or just after of this one, but will leave that for another PR

The infrastructure for this is supposed to be the basis for `sel4test` runs later, but will likely need a bit of refactoring before it can handle all use cases of that (e.g. currently there are only build definitions, but no separate run definitions, and no programatically generated variants yet either). One step at a time..

Some of the info in platforms.yml is unused at the moment (unsupported platforms, machine names). This is mostly there to capture the data that is currently in Bamboo, so we don't need to constantly go back to those definitions. Will purge these when they actually turn out unused in the end.

(This is on top of #95, which should be merged first)

Implements #69

### DIFF
--- a/.github/workflows/deploy-cparser-builder.yml
+++ b/.github/workflows/deploy-cparser-builder.yml
@@ -2,7 +2,7 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 
-# Build and push link-check docker image on changes to relevant files
+# Build and push docker image on changes to relevant files
 
 name: Deploy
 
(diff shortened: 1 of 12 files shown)


lsf37:
(added empty line to fix python style)

xurtis (approved):

lsf37:
(rebased)

PR #104, opened 2021-07-06
make repo-util case insensitive

lsf37:
* GitHub and Bitbucket repo URLs are case insensitive, so some of the manifest project names don't necessarily line up with the full (case sensitive) repo name.
    
  This commit brings repo-util in line with that behaviour so that it returns valid paths for PR GITHUB_REPOSITORY references that might have different case.

* use repo-util in sel4testsim to print repo hashes

* add more info on what `repo` is actually pulling (branch + manifest file). These are harder to recover later and easy to print here.

### DIFF
--- a/scripts/repo-util
+++ b/scripts/repo-util
@@ -12,6 +12,10 @@ manifest itself.
 
 Uses git and repo as shell processes. Hopefully a more stable interface
 than trying to mess with repo python library directly.
+
+Repo URLs are treated as case insensitive to line up with GitHub and Bitbucket
+behaviour. This means, keys in the `projects` dict are lower-case, but we keep
+the spelling for the name as value to reflect what is in the manifest.
 """
 
 import subprocess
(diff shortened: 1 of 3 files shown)


mbrcknl (approved):
LGTM

scripts/repo-util
```diff
@@ -34,7 +34,7 @@ def add_manifest(projects: dict):
         git_cmd = f"git -C {path} config --get remote.origin.url"
         url = subprocess.check_output(git_cmd.split(' '), text=True).rstrip()
         repo = removesuffix(url.split('/')[-1], '.git')
-        projects[repo] = path
+        projects[repo.lower()] = (path, repo)
```
axel-h, line 37: maybe add a comment in the code that lower case name is the preferred sanitized form?

lsf37: yes, good idea. So far the only info on all this is in the commit message.

axel-h (approved):

PR #106, opened 2021-07-07
Deploy C parser images on changes to l4v

lsf37:
The l4v repository has a corresponding action that sends a repository_dispatch event when the C parser changes.

See also seL4/l4v#304


mbrcknl (approved):
Looks good, except I wonder if it introduces some races.

It looks to me like edits in `ci-actions/scripts` or `ci-actions/docker` will trigger a job to build the `cparser-builder` image, and at the same time, a job to build the `cparser-run` image. Then, when the `cparser-builder` job finishes, it triggers another job to build the `cparser-run` image again.

The only way I can see to avoid that is to remove most of the triggers from the `cparser-run` workflow, leaving only the `seL4-platforms/**` and  `cparser-run/**` path triggers. All the other triggers will now be subsumed by the `cparser-build` `respository_dispatch` trigger.

Something I noticed while understanding how this works... IIUC, the purpose of the `cparser-builder` image is only to supply the standalone parser binaries to the `cparser-run` and `preprocess` docker builds? If so, we might be able to make the `cparser-builder` image smaller by throwing out all but the binaries needed for those builds, using a [multi-stage build]
(https://docs.docker.com/develop/develop-images/multistage-build/). I could make a pull request if you think that could work.

lsf37:
> Looks good, except I wonder if it introduces some races.
> 
> It looks to me like edits in `ci-actions/scripts` or `ci-actions/docker` will trigger a job to build the `cparser-builder` image, and at the same time, a job to build the `cparser-run` image. Then, when the `cparser-builder` job finishes, it triggers another job to build the `cparser-run` image again.

I was initially fine with that race, because the last push will overwrite the `cparser-run` that might have used a too-old `coarser-builder`, but see below for more..
 
> Something I noticed while understanding how this works... IIUC, the purpose of the `cparser-builder` image is only to supply the standalone parser binaries to the `cparser-run` and `preprocess` docker builds? If so, we might be able to make the `cparser-builder` image smaller by throwing out all but the binaries needed for those builds, using a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/). I could make a pull request if you think that could work.

Actually, now that you say this, pushing the `cparser-builder` to dockerhub makes no sense -- I should remove that deployment action. It is absolutely intended to be a throw-away container for multi-stage build.

It used to be an inline multi-stage build, but I factored it out, because two separate images need it now (`preprocess` and `cparser-run`). And of course I've forgotten to trigger `preprocess`. All in all, pretty great :)

So, what I think I should do:
 - remove `cparser-build` deploy action
 - trigger `preprocess` and `cparser-run` directly on the `cparser-deploy` event

This is likely to lead to `cparser-build` being built twice, one for each deployment, but that is not really a problem.

lsf37:
Ok, new version pushed (overwriting the old)

lsf37:
I'll merge this now so I can test it by sending events from the l4v side (it's hard to test without it being live). Happy to keep working on this in an additional PR, though, if there are more comments here.

lsf37:
Ok, and that test reveals why I did do that deployment int the first place. The image needs to exist in the registry of course for the other ones to use it by that reference name, unless you explicitly build it locally first, which I guess is what I should be doing. It's not really multi-stage any more, but it's either that or duplication, and the builder is more complex than the other two images, so I don't really like the duplication.

Will experiment with explicitly building the builder in the github action.

mbrcknl:
(I'm having trouble keeping up with you, since I'm in a COVID vaccination queue, so you might already be ahead of me here...)

I guess I would still advocate a separate build job (like before), but with:
- Triggers adjusted to avoid races.
- The `cparser-builder` image minimised using #107.

For the latter, I think it might be nice to use `ubuntu` as the base image for the final stage. Or whatevert he base image for `trustworthysystems/sel4` is. (I haven't found the Dockerfile for that yet.) This would mean we could use `cparser-builder` for  ad-hoc  C  parsing.

lsf37:
Yes, I think I'm happy with that. Have wasted enough time on the github docker action for now. Will put up a new PR for that.

PR #107, opened 2021-07-07
minimise cparser-builder docker image

mbrcknl:
The `cparser-builder` image is only used to provide standalone c-parser
executables to other Docker image builds. This commit minimises the size of the
image that needs to be pushed, by including only the executables needed for
downstream image builds.


mbrcknl:
@lsf37 I'm not sure if this is what we want. If the purpose of this image is just to supply executables image builds for `cparser-run` and the pre-process test, then this is probably fine. If it's also intended to be used as a general-purpose standalone parser, then we would probably want to replace `scratch` with whichever imag//e `trustworthysystems/sel4` is based on. With  only `scratch`, it's not possible to run the C parser from images based on this commit, because shared libraries (for `/c-parser/standalone-parser/*/c-parser`) and the shell (for `/c-parser/standalone-parser/c-parser`) are missing.

I've tested building `cparser-builder` with this change, and also tested running the C parser with `scratch` changed to `ubuntu`. I haven't tested any downstream image builds.

lsf37:
This is what we used to want (and used to have), but I factored out the builder for the `preprocess` action which also needs it. That's why it has its own docker file. I should have triggered that build as well, but forgot about it.

See my comment on #106 for what I think we should do instead.

mbrcknl:
> This is what we used to want (and used to have), but I factored out the builder for the `preprocess` action which also needs it. That's why it has its own docker file. I should have triggered that build as well, but forgot about it.
> 
> See my comment on #106 for what I think we should do instead.

This is still a separate factored-out image build, it just has less stuff in it for the push. But if we change #106 such that we don't push `sel4/cparser-builder` at all, then I agree that there's no point in minimising this image.

I'll have another look at #106, and then probably close this...

lsf37:
> > This is what we used to want (and used to have), but I factored out the builder for the `preprocess` action which also needs it. That's why it has its own docker file. I should have triggered that build as well, but forgot about it.
> > See my comment on #106 for what I think we should do instead.
> 
> This is still a separate factored-out image build, it just has less stuff in it for the push. But if we change #106 such that we don't push `sel4/cparser-builder` at all, then I agree that there's no point in minimising this image.

Right, I didn't look properly. This is better in any case, whether with push the image or not.
 
> I'll have another look at #106, and then probably close this...

I think we should actually merge it :-)

On #106, while my plan works perfectly fine on the command line locally, it doesn't work in the GitHub action yet, which doesn't seem to find the builder image locally. See #108 where I'm still trying things out.

If I can't get it to work with a local build, I might push the builder after all and fall back to the chain you suggested, i.e. send all triggers to the builder, and once that it pushed, trigger the actual builds we want to deploy.

lsf37 (approved):

PR #108, opened 2021-07-07
Explicitly build cparser-builder container first

lsf37:
Since the builder container is not inline, we need to tell docker how to build it.


lsf37:
Abandoning this in favour of pushing the builder image after all.

PR #113, opened 2021-07-12
run proofs on AWS

lsf37:
This implements a GH action similar to `run-proofs` which runs the l4v proofs on AWS instead of on GitHub runners directly. This should remove the resource problems the GitHub runners presented for the proofs.

This also implements a cache for generated Isabelle images on S3.

### DIFF
--- a/aws-proofs/README.md
+++ b/aws-proofs/README.md
@@ -26,6 +26,7 @@ The action expects the following two environment variables to be set.
 
 - `AWS_ACCESS_KEY_ID`: AWS user with sufficient rights to start and stop instances
 - `AWS_SECRET_ACCESS_KEY`: secret access key for that user
+- `AWS_SSH`: secret ssh key for `test-runner` user on AWS instance
 
 ## Example
 
--- a/aws-proofs/README.md
+++ b/aws-proofs/README.md
@@ -1,0 +1,68 @@
+<!--
+  Copyright 2021, Proofcraft Pty Ltd
+
+  SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Run the seL4 proofs on AWS
+
+This action checks out the verification manifest from
+<https://github.com/seL4/verification-manifest>, updates the relevant repository
+to the pull request state, and runs the specified proof images on AWS. It
+automatically spins up and terminates the corresponding AWS instance.
+
+## Arguments
+
+- `L4V_ARCH` (required): which architecture to run the proofs for. One of `ARM`,
+  `ARM_HYP`, `RISCV64`, `X64`
+- `session`: which session to run (e.g. `CRefine` or `ASpec`). Runs all sessions
+  if unset.
+- `isa-branch`: which branch of the Isabelle repo (e.g. `ts-2020`, default as in manifest)
+- `manifest`: which manifest file (e.g. `devel.xml`, `mcs.xml`, `default.xml`)
+
+## Environment
+
+The action expects the following two environment variables to be set.
+
+- `AWS_ACCESS_KEY_ID`: AWS user with sufficient rights to start and stop instances
+- `AWS_SECRET_ACCESS_KEY`: secret access key for that user
+
+## Example
+
+Put this into a `.github/workflows/` yaml file, e.g. `proofs.yml`:
+
+```yaml
+name: Proofs
+
+on: [pull_request_target]
+
+jobs:
+  check:
+    name: Proofs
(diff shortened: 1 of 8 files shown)


lsf37:
Now ready for review (and merge from my side).

The portable shell test fails on the `<<<` operator, which I'd like to keep. (This is an explicit bash script anyway).

aws-proofs/run
```diff
@@ -0,0 +1,42 @@
...
+# Copyright 2021, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# ./run [ci-actions-branch]
+#
+# This script is the entry point on the AWS instance.
+#
+# Currently, the VM image must be updated manually if this script changes.
```
mbrcknl, line 12: Am I right in understanding that you've built an AMI that's just some base distro plus some pre-configuration to make it  suitable for running tests? (e.g. a `test-runner` user, this script, plus some things installed). Is the process of building the AMI documented somewhere? It sounds like you intend to automate building the AMI, but that's still on the TODO list. (It's not a problem, I'm just curious.)

aws-proofs/steps.sh
```diff
@@ -0,0 +1,68 @@
...
+until nc -w5 -z ${IP} 22; do echo "."; sleep 3; done
+
+eval $(ssh-agent)
+ssh-add -q - <<< "${AWS_SSH}"
+
+echo "::endgroup::"
+
+CI_BRANCH=aws-proofs
+
+ssh -o StrictHostKeyChecking=no test-runner@${IP} \
```
mbrcknl, line 51: I guess there's no real harm done if we connect to the wrong server. But if we're building our own AMI, could we pre-load it with known SSH keys, so we don't need `StrictHostKeyChecking=no`? (Not for this pull request, but perhaps for the wish-list.)

aws-proofs/run
```diff
@@ -0,0 +1,42 @@
...
+#
+# It clones the ci-actions repo and runs the test driver from there. Expects to
+# be running in /home/test-runner/ as the test-runner user. Input to the test
+# driver is via environment variables as in the docker steup.
```
mbrcknl, line 16: I like this level of indirection. :+1:

mbrcknl (approved):
Looks good!

Just a couple of very minor nits, and a few questions to satisfy my curiosity.

aws-proofs/README.md
```diff
@@ -0,0 +1,77 @@
...
+on: [pull_request_target]
+
+jobs:
+  check:
+    name: Proofs
+    runs-on: ubuntu-latest
+    strategy:
+          fail-fast: false
+          matrix:
+            arch: [ARM, ARM_HYP, RISCV64, X64]
```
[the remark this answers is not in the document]
lsf37: :+1:

aws-proofs/README.md
```diff
@@ -0,0 +1,77 @@
...
+<https://github.com/seL4/verification-manifest>, updates the relevant repository
+to the pull request state, and runs the specified proof images on AWS. It
+automatically spins up and terminates the corresponding AWS instance.
+
+## Arguments
+
+- `L4V_ARCH` (required): which architecture to run the proofs for. One of `ARM`,
+                 `ARM_HYP`, `RISCV64`, `X64`
+- `session`:     which session to run (e.g. `CRefine` or `ASpec`). Runs all
+                 sessions if unset.
```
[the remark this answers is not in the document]
lsf37: :+1:

aws-proofs/steps.sh
lsf37: The AMI does have a host key set up, but the GitHub runner that connects to it doesn't know that key and it comes up with a different IP every time, of course. I might have a brief look if I can prime the GitHub runner with the host key somehow, because it would also get rid of the warning message that you still get anyway.

lsf37: So the obvious thing does not seem to have worked. I'll look into it a bit more tomorrow in a separate PR.

mbrcknl: It's not really important. There's no sensitive information sent to runner. Worst (and very unlikely) case, somebody pretends to run our tests, and sends back a success when it should have been a failure. But a lot of other things have to go very wrong for someone to be able to *exploit* that. :-)

lsf37: True, it's not critical at all. It's mainly that my curiosity is now piqued on how you would do it :-)

PR #115, opened 2021-07-14
slightly better ssh use

lsf37:
- set up host key before `ssh` (doesn't really add security, but quietens down `ssh`)
- pass env via `ssh` (needs VM sshd config update)

I couldn't find a good method to get the host key from AWS without waiting forever. A new host key is generated at boot time so that not all VMs have the same key. It is echoed on the console during boot, and you can even get the console output from AWS, but only about 3 min after boot on average, and that is too long for my impatience.

So, since no secret data is being passed to the VM and the only output we use from `ssh` is logs and overall test result, I think this is fine.

### DIFF
--- a/aws-proofs/steps.sh
+++ b/aws-proofs/steps.sh
@@ -16,7 +16,7 @@ aws configure set default.region us-east-2
 aws configure set default.output json
 
 echo "Starting AWS instance..."
-aws ec2 run-instances --image-id ami-02109a93a0bf86295 --count 1 \
+aws ec2 run-instances --image-id ami-0a1614a3e5ad44d6b --count 1 \
                       --instance-type c5.4xlarge \
                       --iam-instance-profile "Name=test-runner-role" \
                       --security-group-ids sg-0491b450a86520294 \
--- a/aws-proofs/steps.sh
+++ b/aws-proofs/steps.sh
@@ -44,31 +44,34 @@ until nc -w5 -z ${IP} 22; do echo "."; sleep 3; done
 eval $(ssh-agent)
 ssh-add -q - <<< "${AWS_SSH}"
 
+# Prime GH runner with VM ssh host key. This does not gain any actual security,
+# because we have nothing to verify the key against. It only makes ssh/scp less chatty.
+mkdir -p ~/.ssh
+ssh-keyscan -t ecdsa ${IP} 2> /dev/null >> ~/.ssh/known_hosts
+
 if [ "${GITHUB_EVENT_NAME}" = "pull_request_target" ]
 then
-  GH_REF=${GH_HEAD_SHA}
-else
-  GH_REF=${GITHUB_REF}
+  GITHUB_REF=${GH_HEAD_SHA}
 fi
 
 echo "::endgroup::"
 
-CI_BRANCH=master
+INPUT_CI_BRANCH=master
 
-ssh -o StrictHostKeyChecking=no test-runner@${IP} \
-    "bash -c \"export INPUT_L4V_ARCH=${INPUT_L4V_ARCH}; \
-               export INPUT_MANIFEST=${INPUT_MANIFEST}; \
-               export INPUT_ISA_BRANCH=${INPUT_ISA_BRANCH}; \
-               export INPUT_SESSION=${INPUT_SESSION}; \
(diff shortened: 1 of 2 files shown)


lsf37:
This is still draft until really tested via github on a test repo.

aws-proofs/steps.sh
```diff
@@ -44,30 +44,33 @@ until nc -w5 -z ${IP} 22; do echo "."; sleep 3; done
...
-               export INPUT_SESSION=${INPUT_SESSION}; \
-               export INPUT_CACHE_NAME=${INPUT_CACHE_NAME}; \
-               export INPUT_CACHE_READ=${INPUT_CACHE_READ}; \
-               export INPUT_CACHE_WRITE=${INPUT_CACHE_WRITE}; \
-               export GITHUB_REPOSITORY=${GITHUB_REPOSITORY}; \
-               export GITHUB_REF=${GH_REF}; \
-               export GITHUB_BASE_REF=${GITHUH_BASE_REF}; \
-               export GITHUB_WORKSPACE=/home/test-runner; \
-               ./run ${CI_BRANCH}\""
+INPUT_CI_BRANCH=master
```
lsf37, line 59: This is `INPUT_CI_BRANCH` now because the VM is configured to only accepts `INPUT_*` and `GITHUB*` environment variables.

mbrcknl (approved):
Looks good!

kent-mcleod:
> I couldn't find a good method to get the host key from AWS without waiting forever. A new host key is generated at boot time so that not all VMs have the same key. It is echoed on the console during boot, and you can even get the console output from AWS, but only about 3 min after boot on average, and that is too long for my impatience.

Do you know if you can set it up to use certificate authentication instead of public-key?

lsf37:
> Do you know if you can set it up to use certificate authentication instead of public-key?

I didn't even know what that was :-). I think, yes, this could be set up and it sounds like it would actually achieve security, although I'd have to learn a bit more about how it works. Specifically, what would we use as the certificate authority?

kent-mcleod:
I haven't used it before myself.  I thought there would be something that AWS offers that's already integrated into the available authentication options.  It looks like it's more BYO cert authority though.

mbrcknl:
I'm also new to SSH certificate authentication. It seems nice and very useful in general, but perhaps not so much here.

Since EC2 doesn't seem to provide any built-in support for SSH certificate authorities, I guess the only way to make use of host certificate authentication would be for the client that creates the server to somehow tell the server what host keys and certificate to use. If you're going to do that, then the client might as well just use its knowledge of the host keys to construct an old-fashioned `known_hosts` file.

If we wanted the client to specify server host keys, I guess the way to do it would be using server metadata and a startup script. Seems a bit icky, for two reasons:
1. It would mean that unprivileged processes on the server would be able to see the host keys.
2. I expect there would be a brief period where the server would use the randomly generated host keys before the specified keys are installed.

If we don't care about 1, and make the client to handle 2, this might be quicker than waiting for the console output to appear in the AWS API. Not sure it's worth the hassle, though.

lsf37:
I'm quite happy leaving this as is, since there is almost no impact of a man-in-the-middle attack on this one, but I might take it as an opportunity to learn more about ssh CA. For another day, though :-)

lsf37:
(Will add a temporary commit on top of this for testing the action)

lsf37:
(and rebasing over master)

mbrcknl:
> I'm quite happy leaving this as is, since there is almost no impact of a man-in-the-middle attack on this one, but I might take it as an opportunity to learn more about ssh CA. For another day, though :-)

Yeah, I think this is fine.

mbrcknl:
Haven't finished reading it yet, but possibly relevant for a future iteration:
https://docs.aws.amazon.com/systems-manager/latest/userguide/session-manager.html

lsf37:
Ok, this works now. Will merge.

Happy to keep looking into providing actual host ID, because it looks like it would be nice to report intermediate progress in  the GitHub checks API which needs access to a security token. One easy way to provide that token is through the environment from GitHub, but we could also just the AWS secret manager. So not a big deal if host ID turns out to be too complex, but worth investigating for a bit.

PR #116, opened 2021-07-15
aws-proofs: more robust log collection

lsf37:
The placement of logs changes between Isabelle2019 and Isabelle2020. This commit matches either and also prevents the overall test from failing if there are no logs to be found for some reason.

### DIFF
--- a/aws-proofs/vm-steps.sh
+++ b/aws-proofs/vm-steps.sh
@@ -107,7 +107,10 @@ fi
 # bundle up logs for artifact upload outside the VM
 # xz compression does help even if there are many .gz files in this set
 cd ~/.isabelle/heaps
-tar -Jcf ~/logs.tar.xz */log/*
+# usually one of the two globs below will not exist
+shopt -s nullglob
+# do not fail the test if log collection fails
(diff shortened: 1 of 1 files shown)


mktnk3 (approved):
Nice :)

aws-proofs/vm-steps.sh
```diff
@@ -107,7 +107,10 @@ fi
 # bundle up logs for artifact upload outside the VM
 # xz compression does help even if there are many .gz files in this set
 cd ~/.isabelle/heaps
-tar -Jcf ~/logs.tar.xz */log/*
+# usually one of the two globs below will not exist
+shopt -s nullglob
+# do not fail the test if log collection fails
+tar -Jcf ~/logs.tar.xz */log/* */*/log/* || touch ~/logs.tar.xz
```
mbrcknl, line 113: Do we care that an empty file is not valid input to `xz -d`? Probably not.

mbrcknl (approved):
Nice use of `nullglob`!

aws-proofs/vm-steps.sh
lsf37: you mean for uncompressing? It'll fail, but that is guarded as well on the other end.

PR #123, opened 2021-07-16
sel4test hardware builds

lsf37:
This adds an action for the hardware builds in sel4test, leaving images for artefact upload.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -19,7 +19,8 @@ architectures: [x86, arm, riscv]
 #     simulation_binary: string; optional
 #     march: string; optional
 #     req: string; optional
-#     disabled: optional; (null = false)
+#     disabled: optional; (null = false); no hw test on this platform
+#     no_hw_build: optional; (null = false); no hw build for this platform
 
 platforms:
   SPIKE32:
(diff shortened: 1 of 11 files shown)


seL4-platforms/platforms.yml
```diff
@@ -260,7 +283,7 @@ platforms:
 mcs_unsupported_platforms:
 # tx1 - scheduler accuracy fails, problem with clk rate in kernel or timestamp
 # at user level?
-- TX1
+# - TX1
```
Xaphiosis, line 286: no idea what this commenting out implies

Xaphiosis (approved):
I don't see anything obviously wrong, but I'm not the most qualified reviewer for this...

seL4-platforms/platforms.yml
lsf37: It's hidden in the commit message :-). The not-supported-for-mcs list is from Bamboo, and it mostly checks the list, but it has a special case for some unrelated TX1 stuff that happens to add MCS tests back into the queue. So it's not actually unsupported. To get the test set exactly equal with what is running on Bamboo, I removed it here.

sel4test-hw/action.yml
```diff
@@ -0,0 +1,35 @@
+# Copyright 2021, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+name: 'seL4Test/Hardware Builds'
+description: |
+  Runs sel4test buils for all hardware test platforms.
```
lsf37, line 7: typo: buils -> builds

PR #126, opened 2021-07-20
action for seL4 tutorial tests

lsf37:
Adds a GitHub action for the seL4 tutorial tests. 

- can be filtered by app, test name, and arch
- can dump a full build matrix (turns out to be too large to make much sense on GitHub, but leaving the code here in case it is useful elsewhere later)
- includes some cleanup and refactoring of generic code to simplify the setup of this action
- add colour to logs for sel4test-type logs (like this one); this will take effect of other actions as well


lsf37:
This touches a fair few files, probably easier to review by commit.

kent-mcleod (approved):

PR #129, opened 2021-07-22
add CAmkES test action

lsf37:
includes simulation runs and hardware builds, but not hardware runs yet (blocked on machine queue)


lsf37:
This PR is on top of #128, so will get smaller once that is merged.

kent-mcleod (approved):

lsf37:
Ok, this should finally be it. Rebased and cleaned up a bit.

PR #134, opened 2021-07-30
improvements for aws-proofs

lsf37:
- fix a small bug (would never fall-through to `-x AutoCorresSEL4`)
- add sorry count
- add kernel sloc

The sorry count could still be a bit nicer (show diff), and I might add that in the future, but this is better than none.


lsf37:
bashism check will complain about `nullglob`, but that is unrelated (and wanted)

Xaphiosis (approved):

PR #135, opened 2021-08-05
sel4test hardware runs

lsf37:
This action downloads previously built images and runs them on specified hardware platforms using the machine queue at UNSW.

The boards and the machine queue are still not entirely stable, but the logic is there and things mostly work.

### DIFF
--- a/camkes-vm/build.py
+++ b/camkes-vm/build.py
@@ -32,9 +32,6 @@ def run_build(manifest_dir: str, build: Build):
 
     if plat.arch == 'x86':
         del build.settings['PLATFORM']  # not used for x86 in this test, avoid warning
-        # build.req set by default in builds.yml for x86
-    else:
-        build.req = plat.req or plat.name.lower()
 
     script = [
         ["../init-build.sh"] + build.settings_args(),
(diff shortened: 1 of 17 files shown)


lsf37:
The bashism check fails on the usual command, which should be fine to ignore.

Longer-term this action will probably not be using ssh anyway, or if it does, will run on a separate account. Currently it is using my UNSW login for testing.

lsf37:
There is one last thing I want to test before this is ready for review, which is reducing the time for `ControlPersist`, because I have the feeling GitHub might actually be waiting for those connections to terminate and this is what is sometimes keeping the post action from running when the job is cancelled (the GH post-action timeout seems to be 5m, and `ControlPersist` was longer). Maybe that's not it, but I'm giving it a shot.

lsf37:
Ok, this has also worked, or at least no made anything worse, the run [here](https://github.com/lsf37/seL4/runs/3266711286) was properly cancelled and lock was free'd.

This is now ready for review.

Xaphiosis (approved):

lsf37:
(rebased)

PR #141, opened 2021-08-15
aws-proofs: enable testboard

lsf37:
Together with adding a workflow in the [gh-testboard](https://github.com/seL4/gh-testboard) repo this enables running testboards as before, just triggered from GitHub.

Results should arrive via email for the committer, and be visible on the status of the testboard commit.

(Also removes `fetch-pr.sh` in favour of always using `fetch-branch.sh`, because we don't need history between base and head for the proof test)


lsf37:
The remaining bashism fail is the unrelated `nullglob`

Xaphiosis (approved):

Issue #91, opened 2021-06-21
action for seL4 releases


lsf37:
This includes camkes, l4v, etc.

Currently this is triggered by a nightly run on Bamboo.

This is probably not quite appropriate any more on the GitHub setup (on push to master for specific repos? manually triggered?), and needs some additional thinking on how to avoid race conditions as in the 12.1.0 release, where some repo had new commits before the test finished, and so the release script didn't trigger correctly.

(but otherwise reuse the existing release scripts)

lsf37:
Mostly done for this repo and seL4, but still needs to be deployed in the other main repos, so keeping this open for a bit longer.

kent-mcleod:
> Currently this is triggered by a nightly run on Bamboo.

Apart from the verification-manifests, the other manifest repositories were updated every time a test passed and the nightly triggers were removed because they were too racy I thought.

lsf37:
> Apart from the verification-manifests, the other manifest repositories were updated every time a test passed and the nightly triggers were removed because they were too racy I thought.

That makes sense and is the natural setup for GiHub as well. sel4test and camkes tests currently run that way already and seem to be going fine, so I'll keep with that direction for the rest.

lsf37:
Only one left is the rumprun repo, then I think we have everything the release scripts care about.

Currently, the trigger goes directly from a repo that has changed to the tests for all of the manifests this repo is mentioned in, i.e. they run concurrently instead of in a chain. This seems to be going Ok for now and not be causing too much overload on the machine queue.

I'd be for keeping it this way (unless we run into resource issues), because it avoids a single repo in the chain blocking all other tests.

There is also still the issue we had in the 12.1.0 release that PRs were merged during the release process, so some manifests refer to a different version than others, and the release got stuck because there was not anything consistent to tag. That was resolved manually, but it'd be nice to not have to do that in the future. For 12.1.0 the issue was exacerbated because of networking issues at D61, so the whole thing took much longer than usual. Only way I can see to ensure all release runs see the same state is to not merge PRs just before a release, i.e. have a code freeze period for a few days while everything is churning through.

lsf37:
Ok, all the hardware + deploy actions are done now. seL4 releases should work as before!

PR #145, opened 2021-08-19
action for seL4 release script

lsf37:
This is similar to [l4v-deploy](https://github.com/seL4/ci-actions/tree/master/l4v-deploy), but runs the seL4 release and bump script instead (which is increasingly misnomed `nightly`).

### DIFF
--- a/manifest-deploy/README.md
+++ b/manifest-deploy/README.md
@@ -1,0 +1,65 @@
+<!--
+     Copyright 2021, Proofcraft Pty Ltd
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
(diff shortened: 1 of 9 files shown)


lsf37:
For testing, this currently still points at the `seL4_release` branch `de-bamboo` and only does a dry-run, not a real push.

Bashism test fails on the usual `<<<`. One of these days I shall figure out how to disable a specific check.

Xaphiosis (approved):

lsf37:
This should be the main bit needed for all seL4 release things (#91) from this repo, but it still needs to be actually used in all other repos apart from seL4, where we have seL4/seL4#493.

PR #152, opened 2021-09-08
CAmkES hardware tests

lsf37:
This action runs the images produced by the `camkes-test` action on hardware.

See also seL4/camkes-tool#88


lsf37:
bashism fail for `<<<` expected as usual

Xaphiosis (approved):

PR #154, opened 2021-09-09
add diff_only option for style check

lsf37:
Currently, we are checking the diff on PRs and all files on push, but we have some repositories that are not style-clean, and that we are only updating as we are touching files. In those situations, we will want the `push` style check on the diff, not on everything.


lsf37:
I tucked on a commit for adding a `.stylefiler` to this PR, which is not really the same topic, but trival

Xaphiosis (approved):

PR #155, opened 2021-09-15
remove tests for KZM and armv6

lsf37:
Support for armv6 is removed from the kernel.

See also seL4/seL4#578


lsf37:
[RFC-8] is now accepted.

kent-mcleod (approved):

PR #166, opened 2021-11-04
repeat and skip test runs

lsf37:
Refactor the test-run driver:
    
- a test command can now report that the test should be skipped and be
  marked as skipped. This should make skipped tests more easily visible.
    
- a test command can now report that the test run should be repeated.
  This is intended for hardware runs where the test command can may be
  able to recognize that a board did not boot properly or that output
  on the serial console was garbled.

- allow python functions in addition to shell commands in the test scripts

- mark skipped tests as SKIP
- repeat sel4bench runs on garbled JSON output
- repeat hardware runs on boot failure

### DIFF
--- a/camkes-test/build.py
+++ b/camkes-test/build.py
@@ -74,7 +74,7 @@ def run_build(manifest_dir: str, build: Union[Build, SimBuild]):
     else:
         print(f"Warning: unknown build type for {build.name}")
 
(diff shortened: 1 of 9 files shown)


lsf37:
This still needs more testing, but should be mature enough for review.

Xaphiosis (approved):

PR #167, opened 2021-11-10
turn sel4bench results into web page

lsf37:
Turns json output from the sel4bench-hw action into a web page.

Example output is [here](https://htmlpreview.github.io/?https://raw.githubusercontent.com/seL4/website_pr_hosting/PR_153/localhost/About/Performance/index.html)


lsf37:
Style is still complaining and I just saw that there are still two functions I wanted to factor out. Hang on a minute before reviewing.

lsf37:
Ok, now ready for review.

sel4bench-web/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Content
+
+The entry point is the script [steps.sh].
+
+[Build] and [platform] configurations are in the respective yaml and python
+files in this repository. See these files for config documentation. They are
+shared with the [sel4bench-hw] and sel4bench actions.
+
+The main test driver is [build.py]. The [build.yml][Build] file contains the
```
Xaphiosis, line 25: `build.yml` or `builds.yml`?

sel4bench/build.py
```diff
@@ -126,6 +133,203 @@ def make_runs(builds: List[Build]) -> List[Run]:
...
+        f.write('Time in cycles to invoke a user-level interrupt handler running in the same\n')
+        f.write('address space as the interrupted thread.</li>\n')
+
+        f.write('<li><strong>IPC call</strong>: ')
+        f.write('Time in cycles for invoking a server in a different address space on the same core.</li>\n')
+
+        f.write('<li><strong>IPC reply</strong>: ')
+        f.write('Time in cycles for a server replying to a client in a different address space on\n')
+        f.write('the same core.</li>\n')
+        f.write('</ul>')
```
Xaphiosis, line 231: could use triple-quote blocks to reduce the `f.write(` calls

Xaphiosis (approved):

sel4bench-web/README.md
lsf37: Indeed `builds.yml`

sel4bench/build.py
lsf37: The problem with the triple quotes is that the indent then looks ugly. We could use string concatenation to write a bit less, but I'm not sure it adds much.

PR #170, opened 2021-11-19
generate hw run matrix automatically

### LINKED ISSUE #168: factor out run matrix for sel4test
Currently, the list of hardware platforms to run on is part of the github worklow. This leads to duplication of that list when the workflow is copied to more than one repository.

So, generation of that matrix should be its own action, which can be called from the workflow instead. Ideally this should include all boards that are not disabled.

It'd be good to get an indication somewhere which boards are disabled/skipped. Might have to be done as a separate thing.

lsf37:
This avoids replicating the run matrix in individual workflow files.

Closes #168

### DIFF
--- a/sel4test-hw-matrix/steps.sh
+++ b/sel4test-hw-matrix/steps.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+#
+# Copyright 2021, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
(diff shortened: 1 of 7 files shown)


sel4test-hw-matrix/steps.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
```
Xaphiosis, line 1: I don't remember the details or how I got bitten by this, but there isn't always a `/bin/bash`. Don't really care, if the other scripts depend on it it's better to be consistent than fiddle for unknown future benefit.

Xaphiosis (approved):

sel4test-hw-matrix/steps.sh
lsf37: It's consistent with the other scripts. We do know there is a `/bin/bash` on the github runners (Ubuntu) and there is also one in the seL4 docker containers, so anything that runs in these should be safe.

PR #178, opened 2022-01-19
Update checkbashisms from 2.21.4 to 2.21.7

axel-h:
Debian has dropped `devscripts_2.21.4.tar.xz` from http://ftp.debian.org/debian/pool/main/d/devscripts, there is `devscripts_2.21.7.tar.xz` now.


lsf37:
Thanks for that, looks like there is actually a change in the bashism script this time!

lsf37 (approved):

PR #179, opened 2022-01-19
Cleanup platforms and add comments

axel-h:
Address comments from https://github.com/seL4/ci-actions/pull/177

However, I wonder if not using the "march" parameter unless really necessary is the better option. In the kernel build script we have platform setting also.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -72,7 +72,7 @@ platforms:
     modes: [64]
     platform: qemu-arm-virt
     simulation_binary: qemu-arm-virt
-    march: armv8a
+    march: armv8a # Cortex-A53 is emulated by default
     disabled: true
     no_hw_build: true
 
@@ -208,7 +208,7 @@ platforms:
     platform: rpi3
     req: [rpi3]
     image_platform: bcm2837
-    march: armv8a
+    march: armv8a # ARMv8 platform currently used in AARCH32 mode only.
     disabled: true
 
   RPI4:
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -64,7 +64,7 @@ platforms:
     modes: [64]
     platform: rockpro64
     req: [rockpro64a]
-    march: armv7a
+    march: armv8a
     disabled: true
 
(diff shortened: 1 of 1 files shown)


lsf37:
> However, I wonder if not using the "march" parameter unless really necessary is the better option. In the kernel build script we have platform setting also.

Some of the build matrix filters use `march` to split the matrix, so it has to be known at this level.

seL4-platforms/platforms.yml
```diff
@@ -208,7 +208,7 @@ platforms:
     platform: rpi3
     req: [rpi3]
     image_platform: bcm2837
-    march: armv8a
+    march: armv8a # ARMv8 platform currently used in AARCH32 mode only.
```
lsf37, line 211: This probably means the build scripts will override the setting to `armv7a`. I assume the file here lists `armv8a` because the intention originally was to use it in 64bit mode, but that then didn't happen.

I'd vote for fixing this here, so that build info and actual build are in sync. Nothing particularly bad happens when they are not, but it's confusing and logs etc will be harder to find.

seL4-platforms/platforms.yml
```diff
@@ -260,7 +260,7 @@ platforms:
     arch: arm
     modes: [32]
     platform: allwinner20
-    # march: ?
+    march: armv7a
```
lsf37, line 263: :+1:

lsf37 (approved):

seL4-platforms/platforms.yml
lsf37: (we should check first what the build actually does, one sec)

lsf37: Nope, I'm wrong, the cmake config also sets `armv8a`. So good to leave as is here.

seL4-platforms/platforms.yml
```diff
@@ -64,15 +64,15 @@ platforms:
     modes: [64]
     platform: rockpro64
     req: [rockpro64a]
-    march: armv7a
+    march: armv8a
```
lsf37, line 67: double-checked this and yes, it should be `armv8a`

Issue #185, opened 2022-02-07
Deploy action should not run on forks of the ci-action repos


axel-h:
The `deploy` action should not run on forks of the `ci-action` repo - because this will fail anyway. Currently I'm seeing that this action run when I update `master` in my fork.

lsf37:
Yes, that makes a lot of sense. I can have a look at that.

PR #186, opened 2022-02-07
github: restrict deployment actions to seL4 org

lsf37:
Running docker and other deployments to the seL4 accounts only makes sense from the seL4 org, not from forks of this repository. This commit should remove noise about missing secrets and failed tests from forks.

Implements #185


axel-h (approved):
Looks good to me

PR #191, opened 2022-02-09
pass through PA tokens for private repo access

lsf37:
This enables further actions to be used on private repos.


lsf37:
Some of these already work, but produce warnings because the `token` argument is not declared. This declares those and adds documentation. `thylint` is the one that needed slightly more explicit pass-through.

Xaphiosis (approved):

PR #192, opened 2022-02-14
doc: fix description

axel-h:
fix copy/paste legacy and use a proper desrciption.

### DIFF
--- a/repo-checkout/action.yml
+++ b/repo-checkout/action.yml
@@ -4,12 +4,12 @@
 
 name: 'Repo Checkout'
 description: |
-  Does sel4test test runs from previously built images for all hardware test platforms.
+  Checks out repositories based on a manifest file.
 author: Gerwin Klein <gerwin.klein@proofcraft.systems>
 
 inputs:
   manifest_repo:
-    description: Manifest repostory (e.g. 'sel4test-manifest')
+    description: Manifest repostory (e.g. 'sel4test-manifest', 'sel4bench-manifest')
     required: true
   manifest:
     description: Manifest file


repo-checkout/action.yml
```diff
@@ -4,12 +4,12 @@
...
 name: 'Repo Checkout'
 description: |
-  Does sel4test test runs from previously built images for all hardware test platforms.
+  Checks out repositories based on a manifest file.
 author: Gerwin Klein <gerwin.klein@proofcraft.systems>
 
 inputs:
   manifest_repo:
-    description: Manifest repostory (e.g. 'sel4test-manifest')
+    description: Manifest repostory (e.g. 'sel4test-manifest', 'sel4bench-manifest')
```
lsf37, line 12: Why add sel4bench specifically? There are many more that could be listed here.

lsf37 (approved):
Thanks for spotting that.

repo-checkout/action.yml
axel-h: Some reasons:
- two references are better than one
- both sel4test and sel4bench are a kind of minimum reference for all the rest
... otherwise, nothing specific. Main practical reason might be, that I'm currently working on the sel4bench repo.

lsf37: > - two references are better than one

Strongly disagree with that one :-) Useful values for references are 0, 1, or all. 2 is random.

> both sel4test and sel4bench are a kind of minimum reference for all the rest
> ... otherwise, nothing specific. Main practical reason might be, that I'm currently working on the sel4bench repo.

It's fine to merge it, it's not wrong, I was just wondering, because sel4bench is probably the least used manifest.

lsf37:
Looks like Debian updated `devscripts` again. Fine to ignore here.

axel-h:
Looks like https://deb.debian.org/debian/pool/main/d/devscripts/devscripts_2.21.7.tar.xz is gone and thus the link check fails. There is `devscripts_2.21.7~bpo11+1.tar.xz` (seem it has become subject to "backports" now) or `devscripts_2.22.1.tar.xz`

lsf37:
We'll want `devscripts_2.22.1.tar.xz`, since we're only really extracting `bashims` from it. The main reason I do this manually is that `devscripts` has loads of dependencies which completely bloat the docker container even though we only need this one single script.

axel-h:
rebased on top of https://github.com/seL4/ci-actions/pull/193

PR #194, opened 2022-02-15
repo-checkout: make sure PyGithub is available

lsf37:
So far this action had only been used in deployment workflows where no
pull-request relevant code was running (on push actions only). Since we
are now using it on seL4bench PRs, we need to make sure the
dependencies for the python PR info code are available.

Fixes seL4/sel4bench#24

### DIFF
--- a/repo-checkout/steps.sh
+++ b/repo-checkout/steps.sh
@@ -14,6 +14,8 @@ curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
 chmod a+x ~/bin/repo
 PATH=~/bin:$PATH
 
+pip3 install -U PyGithub
+
 echo "::endgroup::"
 
 echo "::group::Repo checkout"


repo-checkout/steps.sh
```diff
@@ -14,6 +14,8 @@ curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
 chmod a+x ~/bin/repo
 PATH=~/bin:$PATH
 
+pip3 install -U PyGithub
```
axel-h, line 17: The downside of this is, that it always happens on demand now. If there was a container with PyGit and the repo tool, this can be a stable baseline for reproducible builds.

axel-h (approved):
Looks good enough to fix the immediate problem we are facing.

repo-checkout/steps.sh
lsf37: No, this is definitely not worth a docker action. This action is currently just a script -- downloading a container with a Linux install base as large as the standard GitHub runners takes an order of magnitude longer than installing a python package. You could try to scale it down with something like alpine Linux that only has python and repo etc, but that it's own maintenance problems. Only link-checker currently does that because the upstream version has a docker image that was built that way. Even that is pretty slow.

The pure script actions are usually long finished before the docker actions have even pulled down the container. Docker is only useful if the dependencies are hard to manage, like for compiler tool chains etc.

axel-h: It's hard to argue against the speed argument when this has enough weight. Spped might only be solved if there are dedicated runners that have everything cached, but this will not scale for forks. My point really is that reproducible build environment that don't have any external dependency are worth something when trying to air-gap this. But it's questionable for some Github action because some more stuff would be needed then. Note that we are downloading the the Google repo tool there also, and what we get there is absolutely uncontrolled - and the tool tries to update itself (with some certificate check) also when init runs. I'll make a PR for some docker solution that I have locally now, maybe that is useful some day to have both options.

lsf37: I really don't want docker images for lightweight GitHub actions. The whole point of the architecture of the actions in this repo is to make that not necessary and be able to run scripts directly. Docker is like XML, it  introduces more problems than it solves.

PR #196, opened 2022-02-17
repo-checkout: bash script start settings

axel-h:
Ensure the script is started in a mode where errors or undefined variables cause an exit.

### DIFF
--- a/repo-checkout/steps.sh
+++ b/repo-checkout/steps.sh
@@ -1,11 +1,9 @@
-#!/bin/bash
+#!/bin/bash -eu
 #
 # Copyright 2021, Proofcraft Pty Ltd
 #
 # SPDX-License-Identifier: BSD-2-Clause
 #
-
-set -e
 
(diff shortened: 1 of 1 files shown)


lsf37:
Undefined variables shouldn't cause an exit, we're using them in all kinds of places.

axel-h:
But in this script, it seem that all variables must be define, otherwise running this script does not make sense?

lsf37:
I'm not planning to review that per script and action.

The policy (admittedly not written down anywhere, but so far I was the only person writing these) is to have errors on via `set`, to not provide things on the shebang line in case the shell needs to change (I'm surprised the bashism test didnt' flag this), and for all scripts to expect that some variables may be undefined (e.g. `INPUT_*` may or may not exist, as well as any `GITHUB_*`, although I do rely on `GITHUB_REPOSITORY` and sometimes `GITHUB_REF` to exist, depending on context).

Issue #189, opened 2022-02-08
Make "Test with:" support more formats


axel-h:
Follow-up from https://github.com/seL4/ci-actions/issues/184

GitHub URLs are shown in a short format, even is the actual description text contains a longer format. That can be confusing, as the "test with" does not work then, even if thing look right. 

- Add support for `https://github.com/<org>/<repo>/pull/<id>` because GitHub URLs are shown in the short format `<org>/<repo>#<id>` in the GitHUb GUI, but are stored in the description test in the long format.
- When copy/pasting parts from descriptions from other PRs, link can also end up in the format `[<org>/<repo>#<id>](https://github.com/<org>/<repo>/pull/<id>)`, so supporting a format `[<some text>](https://github.com/<org>/<repo>/pull/<id>)` might also be worth adding at a later stage for even more convenience.

PR #197, opened 2022-02-18
get-prs: support URL and short form

axel-h:
Testing PR with other PRs supports "`<org>/<repo>#<id>`" and "`https://github.com/<org>/<repo>/pull/<id>`" now.
Fixes https://github.com/seL4/ci-actions/issues/189

Test run is https://github.com/axel-h/sel4bench/runs/5241461031?check_suite_focus=true from https://github.com/axel-h/sel4bench/pull/4

### DIFF
--- a/scripts/get-prs
+++ b/scripts/get-prs
@@ -34,12 +34,12 @@ if __name__ == '__main__':
 
     with open(os.environ['GITHUB_EVENT_PATH']) as f:
       event = json.load(f)
-    number = event['number']
+    issue_id = int(event['number'])
 
     # get body dynamically so we don't miss updates on test re-run:
     gh = Github()
     repo = gh.get_repo(os.environ['GITHUB_REPOSITORY'])
-    issue = repo.get_issue(number=int(number))
+    issue = repo.get_issue(number=issue_id)
 
     prs = []
     if issue.body:
--- a/scripts/get-prs
+++ b/scripts/get-prs
@@ -1,10 +1,19 @@
 #!/usr/bin/env python3
 #
 # Copyright 2021 Proofcraft Pty Ltd
+# Copyright 2022 Axel Heider (axelheider@gmx.de)
+#
 # SPDX-License-Identifier: BSD-2-Clause
-
+#
+#
 # Parse a PR description body for comments of the form
-#   "Test with org/repo#id ((, | and) org/repo#id)*"
+#   "[T|t]est with[:]? <ref> ((, | and) <ref>)*"
+# where
+#   <ref> = "<org>/<repo>#<id>" | "https://github.com/<org>/<repo>/pull/<id>"
+#
+# For URLs in comments, the Github GUI displays URLs pointing to PRs on Github
+# as "<org>/<repo>#<id>", even if "https://github.com/<org>/<repo>/pull/<id>" is
+# what the parser finds. For convenience, we support both forms.
 #
 # Expects a GitHub event file to be available for the PR number.
 #
@@ -20,18 +29,45 @@ import sys
 
 from github import Github
 
-test_start_re = "[Tt]est with:?"
-repo_ref_re = "([a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+#[0-9]+)"
-pr_re = re.compile(repo_ref_re)
-repo_sep_re = "([, ]+(and)?[ ]*)"
-test_re = re.compile(f"{test_start_re}[ ]*({repo_ref_re}{repo_sep_re}?)+")
+# Using "(?:...)" makes a regex a non-capturing group, which is an atomic
+# expression that operators can be applied to safely.
+def regex_atom(regex):
+    return f"(?:{regex})"
+
+test_start_re = regex_atom("[Tt]est with:?")
+org_re        = regex_atom("[a-zA-Z0-9_-]+")
+repo_re       = regex_atom("[a-zA-Z0-9_-]+")
+pr_id_re      = regex_atom("[0-9]+")
+
+# Any match of "pr_ref" will yield a 3-tupel, where either the first element or
+# the last two elements ("org_re/repo_re" and "pr_id_re") are not empty. That
+# nicely allows determining what PR reference form we have found.
+pr_ref_short_re = regex_atom(f"({org_re}/{repo_re}#{pr_id_re})")
+pr_ref_url_re   = regex_atom(f"https://github.com/({org_re}/{repo_re})/pull/({pr_id_re})")
+pr_ref_re       = regex_atom(f"{pr_ref_short_re}|{pr_ref_url_re}")
+
+# For the reference list, multiple commas or spaces work ("<ref>,  ,, <ref>"),
+# it's practically support of empty items. The Oxford comma ("<ref>, and <ref>")
+# is also supported.
+ref_sep_re           = regex_atom("[, ]+(?:and)?[ ]*")
+pr_list_next_item_re = regex_atom(f"{ref_sep_re}{pr_ref_re}")
+pr_list_re           = regex_atom(f"{pr_ref_re}{pr_list_next_item_re}*")
+
+# Finally, the complete regex can be build. In Python the number of groups that
+# match an expression is fixed. For repeating matches like in the list, only the
+# last match is returned for each group. To work around this, a matching group
+# for the whole reference list is used and then decomposed in a second step.
+test_with_prs_cre = re.compile(f"{test_start_re}[ ]+({pr_list_re})")
+pr_ref_cre        = re.compile(pr_ref_re)
 
 def find_pr_to_test_with(text):
     prs = []
     if text is not None:
         for line in text.splitlines():
-          if test_re.match(line):
-            prs.extend(pr_re.findall(line))
+            m = test_with_prs_cre.match(line)
+            if m:
+                for t in pr_ref_cre.findall(m.group(0)):
+                    prs.append(t[0] if t[0] else f"{t[1]}#{t[2]}")
     return prs
 
 def unit_test():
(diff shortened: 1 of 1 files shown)


scripts/get-prs
```diff
@@ -1,10 +1,26 @@
...
+#   Test with: <repo-list>
+#   test with: <repo-list>
+#   Test with <repo-list>
+#   test with <repo-list>
+# where <repo-list> is of the forms
+#   <repo>[, <repo-list>]
+#   <repo>[ and <repo-list>]
+# and <repo> is in one of the forms
+#   <org>/<repo>#<id>
+#   https://github.com/<org>/<repo>/pull/<id>
```
lsf37, line 19: Too many words. I'd be fine with something like this:
```suggestion
# Parses a PR description body for comments of the form
#   "[T|t]est with[:]? <ref> ((, | and) <ref>)*"
# where 
#   <ref> = "<org>/<repo>#<id>" | "https://github.com/<org>/<repo>/pull/<id>"
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
 
 from github import Github
 
-test_start_re = "[Tt]est with:?"
-repo_ref_re = "([a-zA-Z0-9_-]+/[a-zA-Z0-9_-]+#[0-9]+)"
-pr_re = re.compile(repo_ref_re)
-repo_sep_re = "([, ]+(and)?[ ]*)"
-test_re = re.compile(f"{test_start_re}[ ]*({repo_ref_re}{repo_sep_re}?)+")
+# Start with a few simple regex'es and re-use them to build more complex ones.
```
lsf37, line 39: This should not be a tutorial
```suggestion
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+# Using "(?:...)" makes a regex a non-capturing group, which is basically an
+# atomic expression again that operators can be applied to.
```
lsf37, line 41:
```suggestion
# Using "(?:...)" non-capturing group to make expression atomic for further operators.
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+org_re        = regex_atom("[a-zA-Z0-9_-]+")
+repo_re       = regex_atom("[a-zA-Z0-9_-]+")
+pr_id_re      = regex_atom("[0-9]+")
+
+# Now things get a bit tricky, Both "pr_ref_short_re" and "pr_ref_url_re" are
+# atoms, but they have matching group inside. When or'ing multiple regex'es,
+# each sub-regex contribute the groups, even if they did not match. Thus, any
+# match of "pr_ref" will yield a 3-tupel, where either the first element or the
+# last two elements ("org_re/repo_re" and "pr_id_re") are not empty. That nicely
+# allows determining what PR reference form we have found.
```
lsf37, line 55:
```suggestion
# Any match of "pr_ref" will yield a 3-tupel, where either the first element or the
# last two elements ("org_re/repo_re" and "pr_id_re") are not empty. That nicely
# allows determining what PR reference form we have found.
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
-repo_sep_re = "([, ]+(and)?[ ]*)"
-test_re = re.compile(f"{test_start_re}[ ]*({repo_ref_re}{repo_sep_re}?)+")
+# Start with a few simple regex'es and re-use them to build more complex ones.
+# Using "(?:...)" makes a regex a non-capturing group, which is basically an
+# atomic expression again that operators can be applied to.
+def regex_atom(regex):
+    return f"(?:{regex})"
 
-if __name__ == '__main__':
+whitespace_re = regex_atom("[ \t]")
```
lsf37, line 45: I don't think `[\t]` makes sense here (this is PR descriptions, not text files), and I find `[ ]` much easier to read than `{whitespace_re}` which obscures what is going on in the rest of the expressions. I.e. I think we should remove whitespace_re and use `[ ]` directly instead.

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+pr_ref = regex_atom(f"{pr_ref_short_re}|{pr_ref_url_re}")
+
+# Since a test can run with PRs from multiple repos, we have to support lists
+# also. A hacky way to achieve this is just defining an atom that is a list
+# item with optionally another separator-atom attached - so a list is just a
+# sequence of such items. That is good enough practically.
+sep_comma_re    = regex_atom(f"{whitespace_re}*,{whitespace_re}*")
+sep_and_re      = regex_atom(f"{whitespace_re}+and{whitespace_re}+")
+sep_line_end_re = regex_atom(f"{whitespace_re}*$")
+repo_sep_re     = regex_atom(f"{sep_comma_re}|{sep_and_re}|{sep_line_end_re}")
```
lsf37, line 67: This changes what the current spec says and is not something we want. We do want to allow `, and` as a separator (a lot of people use Oxford comma), and we do want to allow text after the test reference list for potential comments, so we don't want `$`. We also don't really want `<ref> and and and <ref>` (the previous expression allowed multiple commas because I was lazy, but we shouldn't make it worse).

I suggest:
```suggestion
and_re      = regex_atom(f"and ")
ref_sep_re  = regex_atom(f"[, ]+{and_re}?[ ]*")
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+pr_list_item_re = regex_atom(f"{pr_ref}{repo_sep_re}")
```
lsf37, line 68:
```suggestion
pr_list_item_re = regex_atom(f"{pr_ref}{ref_sep_re}?")
```

(`repo_sep_ref` doesn't really fit, we're matching references to PRs or maybe later branches, but not repos)

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+# last two elements ("org_re/repo_re" and "pr_id_re") are not empty. That nicely
+# allows determining what PR reference form we have found.
+pr_ref_short_re = regex_atom(f"({org_re}/{repo_re}#{pr_id_re})")
+pr_ref_url_re = regex_atom(f"https://github.com/({org_re}/{repo_re})/pull/({pr_id_re})")
+pr_ref = regex_atom(f"{pr_ref_short_re}|{pr_ref_url_re}")
+
+# Since a test can run with PRs from multiple repos, we have to support lists
+# also. A hacky way to achieve this is just defining an atom that is a list
+# item with optionally another separator-atom attached - so a list is just a
+# sequence of such items. That is good enough practically.
```
lsf37, line 63: That didn't make much sense to me and I did the original hack :-) Maybe:

```suggestion
# To keep lists of references a simple repetition (`+`), we define `pr_list_item_re` as always potentially followed by a separator.
# This means `Test with <ref>, <ref>, and` will be accepted, and also
# `Test with <ref><ref>` (i.e. no separator), which works out fine in practice.
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+sep_line_end_re = regex_atom(f"{whitespace_re}*$")
+repo_sep_re     = regex_atom(f"{sep_comma_re}|{sep_and_re}|{sep_line_end_re}")
+pr_list_item_re = regex_atom(f"{pr_ref}{repo_sep_re}")
+
+# Finally, the complete regex can be build. The "$" in the end ensures only
+# lines without any trailing data are accepted.
+# Unfortunately, in Python the number of groups that match an expression are
+# static, for repeating matches like in the list, only the last match is
+# returned for each group. To work around this, we have to make "pr_list" a
+# matching group here and then find all matching "pr_ref" in a second step.
```
lsf37, line 75: Python's behaviour here is logical from its re design and implementation, so I don't really agree with the "unfortunately" and "work around". Maybe:
```suggestion
# The final regex. To extract the <ref> list, we make "pr_list" a  matching group
# here and then find all matching "pr_ref" in a second step.
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+test_with_prs_re   = regex_atom(f"[Tt]est with:?{whitespace_re}*({pr_list_item_re}+)$")
```
lsf37, line 76:
```suggestion
test_with_prs_re   = regex_atom(f"[Tt]est with:?[ ]+({pr_list_item_re}+)")
```

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+
+# Build the compiled regex'es that will be used
```
lsf37, line 78:
```suggestion
# Compile the top-level regexes
```

(plural does not generate `'`)

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+        "Test with: \n" \
+        "Test with:  org3/repo3#43 and org4/repo4#44,https://github.com/org5/repo5/pull/45\n" \
+        "-- Trailing list separators are accepted:\n" \
+        "Test with: https://github.com/org6/repo6/pull/46,\n" \
+        "Test with: org7/repo7#47 and \n" \
+        "-- There must be a valid separator\n" \
+        "Test with: org91/repo91#91org92/repo92#92\n" \
+        "Test with: org93/repo93#93 org94/repo94#94\n" \
+        "-- Lines with trailing data are ignored, even if they have valid repos:\n" \
+        "Test with: org95/repo95#95 android org96/repo96#96\n"\
```
lsf37, line 110: This should pass, although the form here is of course exactly the dangerous/potentially confusing one (i.e. things before `android` should match, and things after not, which might confuse some people, but we haven't seen anyone trying anything like that yet).

The intention is for  something like this to pass:

`Test with org/repo#10   (this PR does thing X)`

lsf37 (commented):
Thanks for that. I have lots of minor comments (on comments :-)) and regular expressions, but the main thing is there.

My main requests are less verbose comments that are more to the point and to not change the current behaviour wrt to separators and line ends.

scripts/get-prs
```diff
@@ -20,13 +36,93 @@ import sys
...
+        "Test with: org1/repo1#41\n" \
+        "test with https://github.com/org2/repo2/pull/42\n" \
+        "Test with: \n" \
+        "Test with:  org3/repo3#43 and org4/repo4#44,https://github.com/org5/repo5/pull/45\n" \
+        "-- Trailing list separators are accepted:\n" \
+        "Test with: https://github.com/org6/repo6/pull/46,\n" \
+        "Test with: org7/repo7#47 and \n" \
+        "-- There must be a valid separator\n" \
+        "Test with: org91/repo91#91org92/repo92#92\n" \
+        "Test with: org93/repo93#93 org94/repo94#94\n" \
```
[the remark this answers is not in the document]
axel-h: Do you really want `org91/repo91#91org92/repo92#92` to pass also, so it yields both `org91/repo91#91` and `org92/repo92#92`? Since we match the expression based on a very strict pattern, that would work actually - but I think it'sgood practice to require at least some separator there.

scripts/get-prs
axel-h: It's just for the first round of reviews ;)

scripts/get-prs
axel-h: will rename `repo_sep_ref` to `pr_sep_re`

scripts/get-prs
axel-h: this works now

axel-h:
What do you thing about it now. I've split up things into 3 commits to allow tracking the interesting changes better.

scripts/get-prs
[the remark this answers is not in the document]
lsf37: I wouldn't want anyone to actually write that, but I wouldn't mind if it was matched. AFAICS it would always match correctly, even with the URL extension, because we always have a number at the end.

The main thing I want is support for comments after the expression, so `$` is definitely out.

We could also just do it properly and remove my hack, i.e. have something like `<pr_ref> (<sep>[ ]+<pr_ref>)*` as the expression and a bit of extra handling to return one list in the end.

lsf37:
> What do you thing about it now. I've split up things into 3 commits to allow tracking the interesting changes better.

Sorry, took a while, project work has started in earnest.. Almost there, just the line-end/separator thing to work out.

scripts/get-prs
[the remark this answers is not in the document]
axel-h: > We could also just do it properly and remove my hack

Done it this was. This accepts the "1xx" references, but ignores the "9xx" references then:
```
        "-- A proper separator is required, otherwise it's considered a comment\n" \
        "-- This can become quite odd\n" \
        "Test with: org114/repo114#114org902/repo902#902\n" \
        "Test with: org115/repo115#115 org116/repo116#116org903/repo903#903\n" \
        "Test with: org117/repo117#117andorg904/repo904#904\n" \
        "Test with: org118/repo118#118gliberishorg905/repo905#905\n" \
```

lsf37 (approved):
Looks good, thank you for your patience :-)

PR #201, opened 2022-03-14
add git-credentials-setup action

mbrcknl:
This action uses git-credentials-store to store credentials for HTTPS access to GitHub repositories.

@lsf37 I'm not sure this is a good idea, so I don't mind if you wan to reject it. Compared to an ssh-agent, it has the advantage that in principle, you can use different credentials for different repos (though I concede that we don't actually need that). But it has the disadvantage that the sensitive material is exposed to all subsequent steps. An ssh-agent makes it a bit harder to extract the actual keys (but maybe it's still possible?). 🤷

I'm leaning towards not doing this, but I thought I'd run it past you before I delete it.

### DIFF
--- a/.stylefilter
+++ b/.stylefilter
@@ -6,6 +6,7 @@ aws-proofs/steps.sh
 scripts/setup-hw-ssh.sh
 l4v-deploy/steps.sh
 manifest-deploy/steps.sh
+git-https-credentials-setup/steps.sh
 
 # uses bash nullglob, which we also do want:
 aws-proofs/vm-steps.sh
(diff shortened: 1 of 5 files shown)


lsf37:
I'd prefer using fewer different mechanisms if possible, so adding a new one should come with good motivation why the existing ones are not sufficient. This one looks more like another way to achieve the same (ssh keys being used in other actions already).

It's good to know that we could do this if we needed to, though.

mbrcknl:
> I'd prefer using fewer different mechanisms if possible, so adding a new one should come with good motivation why the existing ones are not sufficient. This one looks more like another way to achieve the same (ssh keys being used in other actions already).
> 
> It's good to know that we could do this if we needed to, though.

Cool, will close.

PR #204, opened 2022-03-21
aws-proofs: namespace exported C graph-lang

mbrcknl:
This simplifies artifact management in the calling workflow, since it
allows multiple graph-lang exports from different jobs to be collected
in a single artifact, while still using GitHub's upload-artifact action.

Sorry for the churn on this! I'm pretty sure this will be the last pull request on this action.


lsf37 (approved):
No worries, looks good.

PR #205, opened 2022-03-29
aws-proofs: run 2 sessions in parallel

lsf37:
It's not entirely clear if this will really speed things up. If the VM has n nominal cores, this will include hyper-threading, so Isabelle will pick n/2 threads to run. With `-j 2` and without further changes to settings, we will now potentially have all threads on the CPU busy, which would result in slower single sessions.

But: the VMs have a fairly high CPU count (16 threads), and most sessions will not be able to use the full 8 cores all of the time, so there should be significant savings from not being blocked on long sessions.

### DIFF
--- a/aws-proofs/vm-steps.sh
+++ b/aws-proofs/vm-steps.sh
@@ -88,9 +88,9 @@ FAIL=0
 L4V_DIR="$PWD/l4v"
 if [ -n "${INPUT_SESSION}" ]
 then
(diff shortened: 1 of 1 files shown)


lsf37:
The background of this is that ARM and ARM_HYP take over 6h now in Isabelle2021-1, and I'm hoping this will reduce the total time to under 6h again.

mbrcknl (approved):
Definitely worth a try.

PR #206, opened 2022-03-29
aws-proofs: remove hard-coded of session exclusion

lsf37:
Excluding or not excluding the large `AutoCorresSEL4` session should happen at the call site. It turns out that we only really want this on the testboard, and all other sessions needed to artificially providea parameter to remove the exclusion.


mbrcknl (approved):
Makes sense.

PR #208, opened 2022-04-06
platforms: add maaxboard

axel-h:
See https://github.com/seL4/seL4/pull/829

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -84,6 +84,14 @@ platforms:
     req: [imx8mq2]
     march: armv8a
 
+  MAAXBOARD
+    arch: arm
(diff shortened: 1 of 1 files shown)


lsf37:
I think the plan was to use the board for VMs as well, so we should probably enable `hyp` configurations for the test, but I haven't followed the PR, so I'm not quite sure if that support exists already.

axel-h:
> I think the plan was to use the board for VMs as well, so we should probably enable `hyp` configurations for the test, but I haven't followed the PR, so I'm not quite sure if that support exists already.

So far the PRs don't mention hypervisor configurations explicitly. I've added a commit for now, let's see what is does.

lsf37 (approved):
This can now be merged.

Issue #203, opened 2022-03-16
adjust sel4bench parameters for hifive


lsf37:
See discussion in https://github.com/seL4/sel4bench/pull/20

`-DHARDWARE=TRUE -DFAULT=TRUE` should be removed for HiFive from the [Performance webpage](https://sel4.systems/About/Performance/) as well as the CI parameters.

PR #215, opened 2022-07-07
sel4bench: remove HARDWARE and FAULT for hifive

lsf37:
See #203 and seL4/sel4bench#20 for background.

### DIFF
--- a/sel4bench/build.py
+++ b/sel4bench/build.py
@@ -27,7 +27,8 @@ import time
 def adjust_build_settings(build: Build):
     del build.settings['BAMBOO']  # not used in this build, avoid warning
 
-    if build.is_smp():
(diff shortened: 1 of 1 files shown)


axel-h (approved):
Looks ok to me. I've read through the threads, maybe adding a comment that captures the essence could be helpful for the long run, so the python code checks can be understood without digging through the PR/issue history.

kent-mcleod (approved):

lsf37:
> Looks ok to me. I've read through the threads, maybe adding a comment that captures the essence could be helpful for the long run, so the python code checks can be understood without digging through the PR/issue history.

Fair enough, I've added the PR discussion link as a comment.

PR #216, opened 2022-07-11
platforms: remove named machines

lsf37:
Removing dead code. The GitHub CI setup turns out not to need these.


seL4-platforms/platforms.yml
```diff
@@ -314,13 +314,3 @@ mcs_unsupported_platforms:
 - MAAXBOARD # To be tested, for now this just copies the setting of IMX8MQ_EVK.
 - ROCKPRO64
 - ODROID_XU4
-
-# named machines
-# list of { req: platform } (re-using req as name)
-machines:
```
kent-mcleod, line 320: I thought these were needed for sel4bench-hw: https://github.com/seL4/ci-actions/tree/master/sel4bench-hw
Originally, the sel4bench hw runs were always tied to the same machine that was named explicitly.

lsf37: It's still tied to machines, but the mapping was just storing info that is already available elsewhere. You can just as well directly use the machine names now, which the code does here:

https://github.com/seL4/ci-actions/blob/1529b3205775a42add8d86add29f96579dbd48c7/sel4bench/build.py#L123-L124

Even if we needed to do somethings specific for specific machines for other platforms in the future it's now easy to figure out what platform something is, because there is more explicit metadata available.

kent-mcleod (approved):

PR #218, opened 2022-07-11
aws-proofs: sorry-count script has moved to l4v

### LINKED ISSUE #213: use l4v sorry-count.sh
The `aws-proofs/sorry-count.sh` script now lives in the `l4v` repository on the `master` branch.

We cannot remove it here yet, because it is not yet available on all `l4v` branches, in particular not on `rt`.

This issue is to remind us to remove the script in this repo once it is present on all active `l4v` branches.

lsf37:
Closes #213


lsf37:
Shell checker fails on removed files, which is being fixed in #219

aws-proofs/vm-steps.sh
```diff
@@ -91,8 +91,14 @@ do_run_tests || FAIL=1
...
 ~/ci-actions/aws-proofs/kernel-sloc.sh
-echo ""
-cd l4v; ~/ci-actions/aws-proofs/sorry-count.sh; cd ..
+
+# sorry-count script lives in l4v repo; guard against branches where it might be gone
+SORRY_COUNT=misc/stats/sorry-count.sh
+if [ -x "l4v/${SORRY_COUNT}" ]
+then
+  echo ""
+  cd l4v; "./${SORRY_COUNT}"; cd ..
```
axel-h, line 100:
```suggestion
  cd l4v; "${SORRY_COUNT}" ; cd ..
```
No sure what shells or environment this has to work with, but is the `./` really needed?

axel-h (approved):

aws-proofs/vm-steps.sh
lsf37: It is definitely better shell style to always anchor the command, otherwise you don't know what you are calling. With the current value of `SORRY_COUNT` it is redundant, but if that value ever changes to a bare command, the shell could be selecting something from the PATH.

PR #219, opened 2022-07-11
style checkers: filter out renamed+deleted files

lsf37:
So far the style checkers were passed deleted and renamed files in the argument list and it was up to the checkers to ignore those. Some do, some done.

So instead, we filter out those files before they get passed in.


axel-h (approved):
looks good to me.

PR #222, opened 2022-10-05
dashboard: script to generate a test dashboard

lsf37:
Currently this is run manually, eventually it should be run automatically on a schedule.

PR for current output at [docsite](https://github.com/seL4/docs/pull/157) at seL4/docs#157


dashboard/gen-dashboard.py
```diff
@@ -0,0 +1,133 @@
...
+            print(f"WARNING: main repo '{repo}' not found")
+            continue
+        print_repo(repo)
+        for workflow in workflow_data[repo]:
+            print_workflow(repo, workflow_data[repo][workflow])
+
+    print("\n-----\n")
+
+    print("## All repositories\n")
+    ordered_repos = [r for r in sorted(workflow_data.keys())]
```
axel-h, line 122: What magic does this do, why can't you just use the list returned by `sorted()` here?

lsf37: Well spotted, that's a left-over from filtering out repos that were already mentioned in main_repos. I decided to duplicate them after all, because I wanted one complete list somewhere. So yes, this can be collapsed just to `sorted(..)`.

axel-h (approved):
Looks good to me.

lsf37:
rebased + squashed

PR #224, opened 2022-10-06
enable MCS tests for more boards

lsf37:
More boards are supported for MCS than the long list of mcs-unsupported might lead one to believe.

This PR enables more boards, and for some disables specific configurations (e.g. SMP + MCS) that are not supported yet. For these, there are open issues to either enable support or make the configuration impossible to select.

See also seL4/seL4#877 for the discussion on which ones to enable/disable.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -18,7 +18,7 @@ architectures: [x86, arm, riscv]
 #     image_platform: name; optional
 #     simulation_binary: string; optional
 #     march: string; optional
-#     req: list of string; optional
+#     req: list of string; optional; names of test boards in machine queue
 #     disabled: optional; (null = false); no hw test on this platform
 #     no_hw_build: optional; (null = false); no hw build for this platform
 
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -299,18 +299,12 @@ platforms:
     simulation_binary: x86
     march: nehalem
 
+# platforms where MCS is wholly unsupported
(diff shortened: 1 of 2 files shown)


lsf37:
(still on draft, because we need to drop [c8be732](https://github.com/seL4/ci-actions/pull/224/commits/c8be7328922deec88ef8e52a11e0012d7ebf098f) from the PR before it can be merged)

seL4-platforms/platforms.yml
```diff
@@ -300,17 +300,9 @@ platforms:
...
-# not ported yet
-- ODROID_X
-- ZYNQMP
-- IMX8MM_EVK
-- IMX8MQ_EVK
-- MAAXBOARD # To be tested, for now this just copies the setting of IMX8MQ_EVK.
-- ROCKPRO64
-- ODROID_XU4
+  # fails SCHED0012
+  - IMX8MM_EVK
```
axel-h, line 304: Seems I lack some understanding how this is supposed to work. But why is this kept here now and the other have moved. Maybe you could add a comment about the background here, so this gets a bit more clear.

lsf37: Will add a comment. The idea is that this list here are the platforms where MCS doesn't work at all. The ones I've moved into build.py are platforms where MCS works for most configuration combinations, but not for some (those are then excluded).

lsf37:
Sqashed, rebased, and have dropped the `testing` commit, so this is now ready for review and merge.

axel-h (approved):
Looks good to me

kent-mcleod (approved):

Issue #225, opened 2022-10-19
update actions to node16


lsf37:
GitHub [has deprecated `node12`](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/) for javascript actions and now wants `node16`.

This means we should update dependencies in the actions in this repo.

It looks like the github core action itself is still on `node12`. Not sure how that will interact.

There doesn't seem to be a cut-off point yet, but it will not be before 2023.

PR #230, opened 2022-11-09
Node16

lsf37:
- upgrade GitHub actions to node16 (see also #225)
- upgrade import versions for `actions/core` and `actions/exec`

The corresponding GitHub dev blog entry is [here](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/).


lsf37:
90% of the change in this PR (everything under nodes_modules) are just the result of running `npm update`.

lsf37:
Since the `node_modules` stuff is 3rd party upstream code, I think we can let the whitespace test fail on that one.

PR #209, opened 2022-04-06
platforms: add qemu-riscv-virt

axel-h:
See https://github.com/seL4/seL4/pull/818

PR #234, opened 2022-11-10
add simulation tests for RISCVVIRT32

axel-h:
This is a follow-up on https://github.com/seL4/ci-actions/pull/209 to also add support for the 32-bit version of qemu-riscv-virt. It's currently still in draft state, as building fail for me locally. However, that might not be a blocker to add it to the action already, so it can be used once the issues are fixed.

see also: seL4/seL4#951

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -50,6 +50,15 @@ platforms:
     disabled: true
     no_hw_build: true
 
+  RISCVVIRT32:
+    arch: riscv
+    modes: [32]
+    platform: qemu-riscv-virt
+    simulation_binary: qemu-riscv-virt32
+    march: rv32imac
+    disabled: true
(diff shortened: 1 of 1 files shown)


lsf37:
This looks fine from my side, but it will be part of the build matrix for simulation immediately when we merge it (unless we filter it out specifically), so we should make sure first that the build work.

axel-h:
The build still fails locally for me due to a non-trivial issue. But I lack time to investigate it, so please don't merge. Will make it a draft again.

Ivan-Velickovic:
seL4test will build/pass with this patch: https://github.com/Ivan-Velickovic/seL4/commit/5bc7ad983bd1d04ccf3c41994a995af85109ca00 (in addition to this PR).

Ivan-Velickovic (approved):

PR #236, opened 2022-11-11
platforms: disable quartz64 hardware build

lsf37:
Currently, the patch set for quartz64 does not compile yet.

See also seL4/seL4#727


Ivan-Velickovic (approved):

Ivan-Velickovic:
It's just seL4test that doesn't build, I think the kernel still builds.

PR #238, opened 2022-11-17
platforms: enable quartz64 hardware build again

axel-h:
Revert 6298c0c8 because timer tests are disabled for quartz64 now in sel4test.

 See also PR seL4/sel4test#83 for the change in sel4test


lsf37 (approved):
Ok, let's give this a shot again.

PR #239, opened 2022-11-17
webserver: additional build platforms

lsf37:
These have not been in CI before, but are listed as supported in the sel4webserver repo.

They will not automatically be built yet, because the build matrix in the sel4webserver repo lists test platforms explicitly, so we can merge here without breaking anything.

### DIFF
--- a/webserver/builds.yml
+++ b/webserver/builds.yml
@@ -8,6 +8,15 @@ default:
   success: 'buildroot login:'
 
 builds:
+
 - odroid_xu4:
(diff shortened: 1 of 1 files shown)


lsf37:
Forgot to mention: the mode can be left out for ODROID_XU4, because it's determined by the platform definition already anyway (same for the other ones).

axel-h (approved):
Well, if this is risk free to merge to prepare the grounds, then do it

PR #240, opened 2022-11-17
add MCS for standalone kernel compile action

lsf37:
To get more coverage, this is not restricted to the verified configs, but just does one plain and one MCS run for each item in the build matrix.


lsf37:
Seems to work well, run time is still under 1 min, i.e. much faster than other PR checks.

axel-h (approved):
Looks good to me, the `-e` catches a build error then and we want both builds to pass eventually.

PR #241, opened 2022-11-20
isabelle-mirror: use old hg for upstream clone

lsf37:
The local hg version is fixed to a fairly old v3 to preserver git hashes produced by the hg-git version that was used when the Isabelle git clone was first produces.

Using `system-hg` now creates a local clone with features that are too new for that old hg to operate on (in particular feature 'safe-share'). Over the network, old hg is still guaranteed to interoperate correctly with recent servers, so we use that for the clone operation.

Thi should fixe the current mirror failures e.g. [here](https://github.com/seL4/ci-actions/actions/runs/3501666603/jobs/5865506976#step:2:888)


mbrcknl (approved):
👍

PR #248, opened 2023-01-18
enable CI tests for NUM_DOMAINS > 1

lsf37:
As @robs-cse found out, the domain scheduler is entirely untested in CI. This PR rectifies that by:

- adding a `domains` variant dimension to the build matrix
- enabling domain tests for sel4test-sim
- enabling domain tests for sel4test-hw

The combination of domains and SMP is disabled globally, all other combinations are possible.

To not increase the overall number of tests too much, we enabled domain tests in simulation for debug builds (because release is not available for riscv), and for release builds in hardware tests.

Only sel4test-sim and sel4test-hw are currently affected, no other builds.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -178,13 +178,24 @@ class Build:
     def can_smp(self) -> bool:
         return self.get_mode() in self.get_platform().smp
 
-    def set_smp(self) -> bool:
+    def set_smp(self):
         if not self.can_smp():
             raise ValidationException("not Build.can_smp()")
         self.settings['SMP'] = "TRUE"
 
     def is_smp(self) -> bool:
         return self.settings.get('SMP') != None
+
+    def can_domains(self) -> bool:
+        return not self.is_smp()
+
+    def set_domains(self):
+        if not self.can_domains():
+            raise ValidationException("not Build.can_domains()")
+        self.settings['DOMAINS'] = "TRUE"
+
+    def is_domains(self) -> bool:
+        return self.settings.get('DOMAINS') != None
 
     def validate(self):
         if not self.get_mode():
@@ -201,6 +212,8 @@ class Build:
             raise ValidationException("not Build.can_release()")
         if self.is_hyp() and not self.can_hyp():
             raise ValidationException("not Build.can_hyp()")
+        if self.is_domains() and not self.can_domains():
+            raise ValidationException("not Build.can_domains()")
 
     def __repr__(self) -> str:
         return \
@@ -731,6 +744,8 @@ def build_for_variant(base_build: Build, variant, filter_fun=lambda x: True) -> 
                 build.set_smp()
             elif feature == 'hyp' and val != '':
                 build.set_hyp()
+            elif feature == 'domains' and val != '':
+                build.set_domains()
             elif feature == 'debug' and val == 'release':
                 build.set_release()
             elif feature == 'debug' and val == 'verification':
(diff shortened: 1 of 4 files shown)


seL4-platforms/builds.py
```diff
@@ -178,14 +178,25 @@ def is_mcs(self) -> bool:
...
+    def can_domains(self) -> bool:
+        return not self.is_smp()
+
+    def set_domains(self):
+        if not self.can_domains():
+            raise ValidationException("not Build.can_domains()")
+        self.settings['DOMAINS'] = "TRUE"
+
+    def is_domains(self) -> bool:
+        return self.settings.get('DOMAINS') != None
```
Xaphiosis, line 198: my memories are that comparison to None should be done via `is None`, but they're hazy

sel4test-sim/builds.yml
```diff
@@ -28,13 +29,20 @@ variants:
 #     smp: ['', SMP]
 #     hyp: ['', hyp]
 #     mcs: ['', MCS]
+#     domains: ['', DOM]
 #     compiler: [gcc, clang]
 #     mode: [32, 64]
 
 
 # only generate builds for platofrms that have `simulation_binary` set
+# only generate builds with DOM set if debug is also set
```
Xaphiosis, line 38: In another commit you have:
```
# run NUM_DOMAINS > 1 tests only on release builds
```
That would mean NUM_DOMAINS > 1 implies release which is not debug? Would that set DOM? Bit unclear on that.

lsf37: in sel4test-sim we pick debug builds, in sel4test-hw we pick release builds. More overall coverage without increasing number of tests too much.

seL4-platforms/builds.py
```diff
@@ -820,16 +820,16 @@ def match_dict(build: Build, f):
...
-                if v != '' and not build.is_mcs():
+                if (v == '') == build.is_mcs():
                     return False
             elif k == 'smp':
-                if v != '' and not build.is_smp():
+                if (v == '') == build.is_smp():
                     return False
             elif k == 'hyp':
-                if v != '' and not build.is_hyp():
+                if (v == '') == build.is_hyp():
```
Xaphiosis, line 829: This isn't really my area, but the previous test said "if there is a value here but it this isn't hyp", and now it says "if there is a value, it must be hyp, otherwise it can't be hyp". That would imply that for all of these, any non-empty value is always an enabling value? I.e. asking for 1 domain implies that domains are enabled?

Xaphiosis (approved):
Approved if my comments turn out to be spurious.

seL4-platforms/builds.py
lsf37: In this case they're going to be equivalent, but you're right that `.. is not None` would be better style here.

Xaphiosis: then again the SMP test above has `!= None` so up to you if it's worth it

seL4-platforms/builds.py
lsf37: Yes, the distinction here is just empty string or any non-empty string.

The test config does not support setting how many domains, it just sets the env variable `DOMAINS` or not. The individual tests then have to figure out what they want to do with that.

seL4-platforms/builds.py
lsf37: I've added a commit that updates all of them.

robs-cse (approved):
Looks good to me - thanks.

PR #256, opened 2023-03-02
Various updates for binary verification

mbrcknl:
(I've lumped several things into one PR, but am happy to split them up if you'd prefer to review separately.)

Related PRs: seL4/l4v#592, seL4/seL4#982.

## l4v-deploy: add a preprocess `manifest` input

This is a first step towards automatic C-compatible manifest updates for MCS. I haven't done the subsequent steps yet, so I can take this out if you'd prefer to wait for that.

## await-remote-artifacts

This will be used by decompilation workflows to download artifacts from the l4v workflow. Previously, this was effectively inlined into the l4v binary workflow. I intend to move that workflow out of l4v and into graph-refine, so there will be PRs on both those repos (though graph-refine will come a little later).

## aws-proofs: Save extra info with C graph-lang

Previously, the aws-proofs `vm-steps.sh` exported the C graph-lang for the workflow to upload as an artifact. This PR slightly generalises that, delegating to a script in l4v to select the files to export. So there will be a PR on l4v to add that script. (The script happens to export more than the C graph-lang, but the idea is that shouldn't matter to `vm-steps.sh`.)


await-remote-artifacts/action.yml
```diff
@@ -0,0 +1,55 @@
...
+    required: true
+    description: |
+      The workflow run ID containing the artifacts.
+
+  artifact-names:
+    required: true
+    description: |
+      A space-separated list of artifact names to wait for.
+      Artifact names should only contain sensible characters,
+      and definitely no whitespace.
```
lsf37, line 26: Could we define this a bit closer to maybe `[0-9a-zA-Z_\-]` or something like that? Then we won't have to argue about sensible with our future selves :-)

aws-proofs/README.md
```diff
@@ -45,7 +45,7 @@ The action expects the following environment variables to be set.
 - `AWS_ACCESS_KEY_ID`: AWS user with sufficient rights to start and stop instances
 - `AWS_SECRET_ACCESS_KEY`: secret access key for that user
 - `AWS_SSH`: secret ssh key for `test-runner` user on AWS instance
-- `GH_HEAD_REF`: the head of the pull request, if a pull request target
+- `GH_HEAD_SHA`: the head of the pull request, if a pull request target
```
lsf37, line 48: Thanks for spotting that one!

l4v-deploy/steps.sh
```diff
@@ -38,12 +46,18 @@ then
   staging-manifest ssh://git@github.com/seL4/verification-manifest.git
   echo "::endgroup::"
 else
+  if [ -n "${INPUT_MANIFEST}" ]; then
+    declare -a MANIFEST_FILE=("--manifest-file" "${INPUT_MANIFEST}")
+  else
+    declare -a MANIFEST_FILE
+  fi
```
lsf37, line 53: This is  a fairly bash-specific way. We can currently assume bash on the Ubuntu runners, so I'm fine with leaving it in, but if you have a POSIX way of achieving the same, we should probably use that.

(I don't remember a good POSIX way of doing this with multiple parameters where `INPUT_MANIFEST` could contain spaces, so no problem if you want to keep it).

lsf37 (approved):
Looking good from my side (minor comments only). We'll have to wait for the GH Actions outage to be over before we can merge.

For the deployment action wrt MCS: happy to take PRs for it now, but not a high priority at this moment (we are making progress in MCS CRefine so it will need to happen at some point).

l4v-deploy/steps.sh
mbrcknl: The only thing I can think of is to inline the `if` with the `seL4-pp` call. With only one such option, it seems ok, though it obviously wouldn't scale to multiple options.

lsf37: Let's keep it for now. It was just in case you had a solution I didn't know about.

lsf37:
(the trailing whitespace is in vendored code only, so we're ignoring the check)

PR #257, opened 2023-03-02
platforms: enable Odroid-C4

Ivan-Velickovic:
The Odroid-C4 has been added to MQ and passes seL4test in default/MCS configurations. Most likely passes more but am yet to test other configurations.


lsf37 (approved):
Happy to try letting it loose on the full set.

PR #258, opened 2023-03-04
bv-trigger: Add action to trigger BV

mbrcknl:
seL4/l4v#595 shows the intended usage.

-------

When run from an l4v proof workflow, checks if there is a
`kernel-builds` artifact, and if so, sends a repository dispatch event
to graph-refine to start a binary verification run.

Previously, a called workflow was used to trigger BV runs. However, a JS
action causes less clutter in the l4v Actions tab. It's also more
efficient, because it does not need to download all artifacts to test
whether the `kernel-builds` artifact is present, and does not always
require spawning a new runner.


mbrcknl:
I've tested this fairly thoroughly on some test repositories. I think it is less intrusive than the called workflow.

lsf37 (approved):
I generally like moving actions into this repo so that they can share infrastructure here and avoid duplication in workflows.

But: the actions in here are supposed to be mostly shell and python so that they are easy to understand/extend, share infrastructure, and can potentially be ported to other CI systems. (And don't need to draw in loads of JS dependencies every time).

It looks like this one could fairly easily be done with `pygithub`.

There are higher priorities right now, so I'd be fine with merging this and opening an issue to change it later when the rest works.

mbrcknl:
> It looks like this one could fairly easily be done with `pygithub`.

As far as I know, the only official API for fetching artifacts in the current workflow run is the JS actions toolkit. I can't find it now, but I recall reading that the toolkit uses internal APIs whose only public documentation is the JS toolkit code. I think I also read that they are currently reworking some of the internals before making a public API. But for now, the artifact download API used by pygithub will only fetch artifacts for workflow runs that have finished.

I'll merge this for now, and maybe a new API will appear soon.

PR #260, opened 2023-03-08
aws-proofs: update repo on start-up

lsf37:
Update the repo tool when the job starts up.

The repo tool wants to update itself more frequently than the VM images and produces annoying chatter when it is not on the latest version.

### DIFF
--- a/aws-proofs/vm-steps.sh
+++ b/aws-proofs/vm-steps.sh
@@ -10,6 +10,8 @@
 set -e
 
 echo "::group::Setting up"
+curl https://storage.googleapis.com/git-repo-downloads/repo > ~/.local/bin
+chmod 755 ~/.local/bin/repo
 export PATH=/home/test-runner/ci-actions/scripts:/home/test-runner/.local/bin:/home/test-runner/.bin:$PATH
 mkdir ver
 cd ver


aws-proofs/vm-steps.sh
```diff
@@ -10,6 +10,8 @@
 set -e
 
 echo "::group::Setting up"
+curl https://storage.googleapis.com/git-repo-downloads/repo > ~/.local/bin
```
axel-h, line 13: What is the worst thing that could happen if the binary we download here is malicious? There is an implicit TLS certificate check here, but the binary could be anything.

lsf37: We already do the same download in several other actions, this just adds one more instance.

The worst that can happen is similar to the worst that can happen for a PR, because a malicious PR can execute arbitrary code on the test runner. In this case the test runner is on AWS, not GitHub, so it costs money and it can consume compute resources up to the timeout (the timeout is controlled externally outside the VM). It can access the internet, but it does not have any credentials to access anything that is not already public.

Xaphiosis (approved):
Yes please, that update-begging is super annoying in logs.

PR #263, opened 2023-03-16
platfomrms: enable all configs for imx8mq

lsf37:


Ivan-Velickovic (approved):
Not expecting everything to pass but would be good to see what does not.

PR #264, opened 2023-03-17
Fix YAML parsing when `mcs_unsupported_platforms` is empty

Ivan-Velickovic:
This checks whether the `mcs_unsupported_platforms` list in `seL4-platforms/platforms.yml` is empty before trying to iterate over it to avoid errors observed in the following output https://github.com/seL4/seL4/actions/runs/4443812105/jobs/7801440632. I unfortunately did not catch this before approving https://github.com/seL4/ci-actions/pull/263.


lsf37 (approved):
I've tweaked the commit a little to turn it into an empty list so that other iteration code in other modules doesn't need to be touched.

PR #269, opened 2023-05-01
aws-proofs: scale timeouts by 50% for cloud

lsf37:
Scale timeouts by 50% to provide an extra stretch factor for cloud tests.

AWS runners don't seem to be reliable in load, specifically the X64 CRefine session is sometimes timing out even though on a normal run there'd be over 1h left. Scaling the timeout should avoid these.


michaelmcinerney (approved):
👍

corlewis (approved):

PR #271, opened 2023-05-03
sel4bench-web: use monospace font for build command

Ivan-Velickovic:
Very trivial style fix. I think it's better for build commands listed on https://sel4.systems/About/Performance/ are in a monospace font.


lsf37:
I agree with the idea, but we should probably put that in CSS instead. Let me check where those classes are defined.

Ivan-Velickovic:
> I agree with the idea, but we should probably put that in CSS instead. Let me check where those classes are defined.

Yes, I thought that there wasn't a CSS file but I can see the other HTML tags using `class=` so yes this style change should be in the CSS.

lsf37:
Found it: the css definitions are in the seL4/website repo in https://github.com/seL4/website/blob/master/content/css/sel4.css

Ivan-Velickovic:
Depends on https://github.com/seL4/website/pull/271

lsf37 (approved):

PR #278, opened 2023-06-02
platforms: disable RPi4

Ivan-Velickovic:
Cannot successfully run seL4test in release mode, only debug mode works currently. Disable until I figure out the issue.


lsf37 (approved):
We could leave only the debug builds in if you want, but also happy to just switch it off for now until we have figured it out.

PR #279, opened 2023-07-18
scripts: update ts server ID

lsf37:
The ssh ID of the TS login server has changed, so actions currently can't access the machine queue.


lsf37:
(the link check failure is unrelated to this PR)

PR #280, opened 2023-07-18
Add new host keys

wom-bat:
login.trustworthy.systems is now replicated, and uses newer sshd.  Host keys have been regenerated;
reflect that here.

The increase in number of keys made it advisable to clean up the way known_hosts was generated; did the same for .ssh/config for consistency and readability.

### DIFF
--- a/scripts/setup-hw-ssh.sh
+++ b/scripts/setup-hw-ssh.sh
@@ -8,23 +8,32 @@
 # set up ssh for access to UNSW machine queue
 
 mkdir -p ~/.ssh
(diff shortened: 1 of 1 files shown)


wom-bat:
Surprised the linkchecker is run --- there's no HTML here.

lsf37:
> Surprised the linkchecker is run --- there's no HTML here.

It's ok, it's run on globally because the repo is small. The failure is unrelated, I'll need to update the dev scripts link (which is always a good reminder to update the bashisms script itself).

lsf37 (approved):
Thanks!

PR #283, opened 2023-08-22
platforms: add Pine64 Star64

Ivan-Velickovic:
For once all the Star64 PRs are merged.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -97,6 +97,14 @@ platforms:
     march: rv64imac
     req: [p550a]
 
+  STAR64:
+    arch: riscv
+    modes: [64]
+    smp: [64]
+    platform: star64
+    march: rv64imac
+    req: [star64]
+
   SABRE:
     arch: arm
     modes: [32]


seL4-platforms/platforms.yml
```diff
@@ -58,6 +58,14 @@ platforms:
     req: [hifive] # , hifive1]
     march: rv64imac
 
+  STAR64:
+    arch: riscv
+    modes: [64]
+    smp: [64]
+    platform: star64
+    req: [star64]
+    march: rv64imac
```
Ivan-Velickovic, line 67: Not sure what `march` does exactly here, `march: rv64gcb` would be more accurate for the Star64. The only use of `march` I can see is here https://github.com/seL4/ci-actions/blob/607634d9c99af003055024f8bd11c47b4bdc757e/seL4-platforms/platforms.py#L183-L192, where it seems to be ignoring `march` anyways.

lsf37: I might be wrong, but I think the only remaining use of `march` is for grouping in CI matrices. We could try to eliminate it and see what happens..

lsf37 (approved):
👍 Happy for you to merge once the other PRs are done.

(Ok to ignore the devscript link failure for this PR)

seL4-platforms/platforms.yml
lsf37: Did a [survey of this](https://github.com/seL4/ci-actions/pull/385#issuecomment-2712001246) for PR #385 and it's a bit of work. So for now, to merge new platforms we should use the old `rv64imac` even if that is not accurate any more.

seL4-platforms/platforms.yml
```diff
@@ -75,6 +75,14 @@ platforms:
     march: rv64imac
     no_hw_test: true
 
+  STAR64:
+    arch: riscv
+    modes: [64]
+    smp: [64]
+    platform: star64
+    req: [star64]
+    march: rv64imafdc
```
lsf37, line 84: For now we still need the old version

```suggestion
    march: rv64imac
```

PR #291, opened 2023-12-08
platforms: add tqma8xqp1gb to hw test

lsf37:
The corresponding tqma board is now available in the machine queue.


lsf37:
We should also add the board to https://docs.sel4.systems/Hardware/ -- do we have installation instructions like for https://docs.sel4.systems/Hardware/sabreLite.html or similar?

Indanz (approved):

PR #293, opened 2023-12-10
aws-proofs: provide optional NUM_DOMAINS override

lsf37:
If the NUM_DOMAINS input is set, the proof build system will override the KernelNumDomains setting. We make the NUM_DOMAINS value part of the cache key so that we get fast proofs on master for things that already had a successful test on a pull request.


Xaphiosis:
see my question on https://github.com/seL4/l4v/pull/701 ... once that's cleared up, I'll approve both.
Might get a bit expensive testing master twice on every commit, but it's probably the right way to do it for now.

Xaphiosis (approved):

lsf37:
> see my question on [seL4/l4v#701](https://github.com/seL4/l4v/pull/701) ... once that's cleared up, I'll approve both. Might get a bit expensive testing master twice on every commit, but it's probably the right way to do it for now.

Yeah, we'll see if the spend goes up too much we might need to switch to weekly. Let's see how it goes.

PR #297, opened 2024-01-05
support LLVM for RISCV64 kernel stand alone build

axel-h:
See https://github.com/seL4/seL4/pull/1154 for a test run

### DIFF
--- a/standalone-kernel/compile_kernel.sh
+++ b/standalone-kernel/compile_kernel.sh
@@ -40,7 +40,17 @@ case $INPUT_ARCH in
         esac
         ;;
     RISCV64)
-        extra_config="${extra_config} -DRISCV64=TRUE"
+        case $INPUT_COMPILER in
+            gcc)
+                extra_config="${extra_config} -DRISCV64=TRUE"
+                ;;
+            llvm)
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Looking good. And I see the test worked as well. Nice.

We had previously excluded clang for RISCV, because there were some problems. Have these been fixed or is it just that more recent compiler versions now work?

axel-h:
No sure if somebody was actively tracking this. LLVM has been working for me for some time now.

lsf37:
Cool, we should try switching on clang for more of the RISCV tests then.

axel-h:
What I'm still not sure about is what clang triple we prefer and why we do this: `riscv64-unknown-elf` or `riscv64-linux-gnu`

lsf37:
For the pure compilation check I don't think it actually matters for the kernel, because we're not linking with any libraries anyway. I guess that makes the more "honest" one `riscv64-unknown-elf`

axel-h:
True. Wwe no longer have `riscv64-linux-gnu` anyway. I was just fooled again by some other scripts that still contain this, but it is completely unused and thus there was never an issues. More PRs to come...

PR #299, opened 2024-01-09
clang is supported everywhere now

axel-h:
This is following up on the comment https://github.com/seL4/ci-actions/pull/297#issuecomment-1881904182, we can use clang everywhere as the builds seem to work fine. So remove the blockers here, they should be put into specific workflow the in case there are still issues.

Test run is here: https://github.com/axel-h/seL4/actions/runs/7464262889/job/20310849733?pr=137 (added dummy for RV32;  removed some HW platforms so the build does not take ages)

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -147,16 +147,7 @@ class Build:
         return self.settings.get("ARM_HYP") is not None or \
             self.settings.get("KernelVTX") is not None
 
-    def can_clang(self) -> bool:
-        # clang 8 does not support riscv, and Bamboo has no clang for TX1 64:
-        return not (
-            self.get_platform().arch == 'riscv' or
-            self.get_platform().name == 'TX1' and self.get_mode() == 64
-        )
-
     def set_clang(self):
-        if not self.can_clang():
-            raise ValidationException("not Build.can_clang()")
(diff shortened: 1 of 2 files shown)


lsf37 (approved):
Cool, looks good.

Thanks for providing the test runs. I've checked the plain base images for hifive and tx1 on the machine queue and they appear to work without problems, so I'm happy to let this loose on the bigger test runs and see if they all work.

lsf37:
Ok, so there is one clang issue remaining after all: on TX1, the combination of SMP+HYP+MCS fails consistently with clang (and do succeed for gcc).

See e.g. https://github.com/seL4/seL4/actions/runs/7468884979/job/20367008152?pr=1156

They are getting suck (timing out) on
```
<testcase classname="sel4test" name="SCHED_CONTEXT_0014">
  Running test SCHED_CONTEXT_0014 (Test deleting a scheduling context while it runs out of budget on another core)
```

Unless you have immediate fix, I'd exclude that combination from the test for now and open an issue for it.

axel-h:
I don't have a fix here.

lsf37:
Ok, see #300 now. Turns out this issue already exists for the TX2 and ODROID_C4.

PR #303, opened 2024-01-11
make flags more intuitive

axel-h:
- `disabled` -> `no_hw_test`
- `simulation_binary` -> `has_simulation` as boolean

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -16,47 +16,47 @@ architectures: [x86, arm, riscv]
 #     aarch_hyp: list of modes; optional (null = empty)
 #     platform: name; required
 #     image_platform: name; optional
-#     simulation_binary: string; optional
+#     has_simulation: boolean; optional; (null = false); platform has a simulation
 #     march: string; optional
 #     req: list of string; optional; names of test boards in machine queue
-#     disabled: optional; (null = false); no hw test on this platform
-#     no_hw_build: optional; (null = false); no hw build for this platform
+#     no_hw_test: optional; (null = false); no hw test on this platform
+#     no_hw_build: optional; (null = false); no hw build for this platform (implies no_hw_test)
 
 platforms:
   SPIKE32:
     arch: riscv
     modes: [32]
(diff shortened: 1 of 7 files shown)


sel4test-sim/builds.yml
```diff
@@ -65,7 +65,7 @@ build-filter:
 #      - arch: [x86, riscv]
 #        compiler: [gcc]
 #        mode: [32]
-#        simulation_binary:
+#        has_simulation:
```
lsf37, line 68: To make it clearer, let's bring the example in sync with the use above.

```suggestion
#        has_simulation: true
```

seL4-platforms/platforms.yml
```diff
@@ -16,47 +16,47 @@ architectures: [x86, arm, riscv]
...
 #     platform: name; required
 #     image_platform: name; optional
-#     simulation_binary: string; optional
+#     has_simulation: boolean; optional; (null = false); platform is a simulation
 #     march: string; optional
 #     req: list of string; optional; names of test boards in machine queue
-#     disabled: optional; (null = false); no hw test on this platform
-#     no_hw_build: optional; (null = false); no hw build for this platform
+#     no_hw_test: optional; (null = false); no hw test on this platform
+#     no_hw_build: optional; (null = false); no hw build for this platform (implies no_hw_test then)
```
lsf37, line 23:
```suggestion
#     no_hw_build: optional; (null = false); no hw build for this platform (implies no_hw_test)
```

lsf37 (approved):
I think this makes sense. The `simulation_binary` setting is a relic from very old times and at best confusing, and `no_hw_test` is better in line with the rest of the terminology and the labels we use on repos.

@june-andronick if you're looking into what is tested, this PR will change the names of the keys you have to check for.

sel4test-sim/builds.yml
axel-h: done

seL4-platforms/platforms.yml
axel-h: done

PR #305, opened 2024-01-12
platforms: add MaaXBoard

Ivan-Velickovic:
The MaaXBoard is the same SoC as the i.MX8MQ and so all the test configurations that work for the i.MX8MQ-EVK should also work.

Draft because the board is currently having network issues but should be resolved in a couple of days when we get a new network switch.


lsf37 (approved):
Ready to merge whenever the board is.

PR #307, opened 2024-01-16
sel4test-hw: skip riscv+clang+smp+release

lsf37:
Until https://github.com/seL4/seL4/issues/1160 is resolved.

### DIFF
--- a/sel4test-hw/build.py
+++ b/sel4test-hw/build.py
@@ -89,6 +89,11 @@ def build_filter(build: Build) -> bool:
         if build.is_hyp() and (build.is_smp() or build.is_verification()):
             return False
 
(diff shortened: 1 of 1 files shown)


sel4test-hw/build.py
```diff
@@ -89,6 +89,11 @@ def build_filter(build: Build) -> bool:
         if build.is_hyp() and (build.is_smp() or build.is_verification()):
             return False
 
+    if plat.arch == 'riscv':
```
axel-h, line 92: Let's make this platform specific and not arch generic? At least we should see them how the Star64 board works once it's in.

Ivan-Velickovic (approved):

PR #310, opened 2024-01-18
platforms: allow setting extra flags to be passed to build

Ivan-Velickovic:
Untested, not sure how to test (I guess its Python so you never know until it is too late :) )

### DIFF
--- a/seL4-platforms/platforms.py
+++ b/seL4-platforms/platforms.py
@@ -57,6 +57,7 @@ class Platform:
         self.req = None
         self.no_hw_test = False
         self.no_hw_build = False
+        self.settings = {}
         self.__dict__.update(**entries)
         if not self.validate():
             raise ValidationException(f"Platform {name} validation")
@@ -109,6 +110,7 @@ class Platform:
             f"    req: {self.req}",
             f"    no_hw_test: {self.no_hw_test}",
             f"    no_hw_build: {self.no_hw_build}",
+            f"    settings: {self.settings}",
             "  }"
         ]])
         return result.getvalue()
@@ -157,6 +159,9 @@ class Platform:
 
     def cmake_toolchain_setting(self, mode: int) -> str:
         return self.toolchain_arch_str() + str(mode)
+
+    def settings(self) -> str:
+        return self.settings
 
     def get_image_platform(self, mode: int) -> str:
         return self.image_platform or self.get_platform(mode)
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -224,6 +224,10 @@ platforms:
     platform: zynqmp
     req: [zcu102_2]
     march: armv8a
+    # The ZCU102 in machine queue expects a binary image
+    settings:
+        Sel4testAllowSettingsOverride: true
+        ElfloaderImage: "binary"
 
   ZYNQMP106:
     arch: arm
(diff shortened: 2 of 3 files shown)


seL4-platforms/platforms.py
```diff
@@ -109,6 +110,7 @@ def __repr__(self):
             f"    req: {self.req}",
             f"    disabled: {self.disabled}",
             f"    no_hw_build: {self.no_hw_build}",
+            f"    extra_settings: {self.extra_settings}",
```
axel-h, line 113: This will print
```
extra_settings: {'foo': 'foo1', 'bar': 'bar1'}
```
or
```
extra_settings: {}
```
but I assume we want YAML format:
```
extra_settings:
  foo: foo1
   bar: bar1
```

seL4-platforms/platforms.yml
```diff
@@ -118,6 +118,9 @@ platforms:
     platform: maaxboard
     req: [maaxboard1]
     march: armv8a
+    extra_settings:
```
axel-h, line 121: Maybe we should pick a more specific name like `build_params` here. It might also be worth having a note here why these setting are needed and the defaults are not used. Would it make sense to change the defaults instead. This platform is quite new, so there might no be too much dependencies on the default.

seL4-platforms/platforms.yml
```diff
@@ -118,6 +118,10 @@ platforms:
     platform: maaxboard
     req: [maaxboard1]
     march: armv8a
+    # The MaaxBoard in machine queue expects a binary image
+    extra_build_params:
```
axel-h, line 122: We have "settings" already in some other YAML files, so we could also use this name here also?

lsf37: Yes, good idea, that should actually work.

Ivan-Velickovic: Is it good now?

seL4-platforms/platforms.py
axel-h: This is still blocking the merge, unless we don't need YAML output here?

axel-h: Seems we don't need YAML.

axel-h (approved):
Looks good to me.

lsf37 (approved):
Cool. I like it with reusing `settings`.

Ivan-Velickovic:
I gave this a test to make sure sel4test ran after merging and to my surprise, adding `Sel4testAllowSettingsOverride` actually does more than I expected.

For example I ran sel4test with `../init-build.sh -DPLATFORM=zynqmp -DSel4testAllowSettingsOverride=1` and `./init-build.sh -DPLATFORM=zynqmp` which I thought would produced the exact same image since no settings are actually being overridden, but it looks like for some reason that the addition of `Sel4testAllowSettingsOverride` now causes sel4test to default to release mode. Not sure if this will be a problem with CI if debug mode is explicitly being turned on.

lsf37:
Hm, it looks like that might influence a whole bunch of settings:
https://github.com/seL4/sel4test/blob/0a488b92bc7a08acc14d49524d1f22656f286627/settings.cmake#L43-L134

That setting is pretty misnamed, it's not an override, but prevents any default settings from applying. This is probably not really what we want to use. We might need to add an actual override that allows you to set things in addition to the defaults.

PR #314, opened 2024-01-24
camkes-vm: add vm_minimal_zcu102 run

lsf37:
- fix typo in `success` key string in CAmkES VM runs (was `succes`)
- add mechanism to override platform name for VM builds (because `zynqmp` needs `zcu102` in `-DPLATFORM=..`)
- add `vm_minimal_zcu102`

(replaces #313)

### DIFF
--- a/camkes-vm/build.py
+++ b/camkes-vm/build.py
@@ -33,6 +33,10 @@ def run_build(manifest_dir: str, build: Build):
     if plat.arch == 'x86':
         del build.settings['PLATFORM']  # not used for x86 in this test, avoid warning
 
+    # if vm_platform is set, the init-build.sh script expects a different platform name.
+    if build.vm_platform:
+        build.settings['PLATFORM'] = build.vm_platform
+
     script = [
         ["../init-build.sh"] + build.settings_args(),
         ["ninja"],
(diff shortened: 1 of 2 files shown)


lsf37:
This should work without further changes in other repos.

It would be nicer to call the `zcu102` target `zynqmp` in the CMake setup in the CAmkES repos, because the name mismatch with the images that are produced might trip up people in the future, but that can be a separate discussion. Let's first get this merged so the image is tested.

chrisguikema (approved):
Thanks, Gerwin! These changes look good

camkes-vm/build.py
```diff
@@ -33,6 +33,10 @@ def run_build(manifest_dir: str, build: Build):
     if plat.arch == 'x86':
         del build.settings['PLATFORM']  # not used for x86 in this test, avoid warning
 
+    # if vm_platform is set, the init-build.sh script expects a different platform name.
+    if build.vm_platform:
```
axel-h, line 37: Does this require any update the `Build` class in `seL4-platforms/builds.py` then also? Seems any attributes from the YAML file are put into the build object, but there is no check if this potentially clashes somewhere?

PR #316, opened 2024-01-26
camkes-vm: fix typo in settings key string

axel-h:
This fixes another typo, which seems to be the reasons some CAmkES VM hardware test fail - they just don't run what they are supposed to run. We have a successful test now: https://github.com/seL4/camkes-vm-examples/actions/runs/7671627886/job/20910564830?pr=59

### DIFF
--- a/camkes-vm/builds.yml
+++ b/camkes-vm/builds.yml
@@ -57,12 +57,12 @@ builds:
 - vm_minimal_smp_tx1:
     app: vm_minimal
     platform: TX1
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Wow, thanks for that. I wonder why this ever worked, actually. 

Maybe some checking for the presence of unknown keys would be good after all (then this would have shown up). We'd need something that `builds.py` implements as a base and every check can extend with its own keys.

PR #317, opened 2024-01-27
camkes-vm: add vm_minimal for ODROID_C2

axel-h:
Seems the hardware run for ODROID_C2 passes also, so we can add this

See https://github.com/seL4/camkes-vm-examples/actions/runs/7676645400/job/20924491718?pr=59 where I have it as 
`vm_minimal_ODROID_C2`


lsf37 (approved):
Yup, should be in CI. Will resolve the conflict with #318 and merge.

PR #318, opened 2024-01-27
camkes-vm: use platform specific names

axel-h:
It improves the readability of the logs a bit, is we use the platform name in capital letters. That is ssimilar to hat we do for the kernel. Also use it even if there is only one board active, so it is a bit more clear which platform this run on.

Is anything really attached to this names at the moment?


lsf37 (approved):
They are easier to spot. AFAICS, nothing is attached to these names. Let's try it.

PR #321, opened 2024-01-28
Retire the unused pr-queue action

lsf37:
We're not using it on the seL4 repos any more, it never really took off.

Removing it now, because it adds maintenance burden for node.js dependencies.


lsf37:
(need to merge this before #322)

mbrcknl (approved):
:+1:

Issue #282, opened 2023-08-11
Build seL4 and seL4test for all supported platforms


Ivan-Velickovic:
I believe the current status of the CI (for sel4test at least) is that not all platforms that claim to be supported are tested. Obviously we cannot buy all the hardware to actually run the tests, but just compiling them to eliminate any build errors is trivial.

Here's what I've seen missing so far:
* QEMU RISC-V virt 32-bit
* Polarfire
* Rocketchip
* Rocketchip based on ZCU102 FPGA
* Ariane
* Ultra96v2
* TK1-SOM (not sure what this is, the docs page is missing instructions for building sel4test)
* Raspberry Pi 3B
* Raspberry Pi 4B

lsf37:
Agreed. At least compiling does definitely male sense, yes.

lsf37:
This *should* be relatively easy to achieve and should only need the corresponding entries in the platforms.yml file.

lsf37:
Actually, RPI3 and RPI4 are in the test and have been built regularly for quite a while. Have you checked the build logs for the rest of the list? (or the [platforms.yml](https://github.com/seL4/ci-actions/blob/master/seL4-platforms/platforms.yml) file -- `disabled` means no hardware test, but still build. Only `no_hw_build` excludes a platform from the build).

Ivan-Velickovic:
Can you point a run where they were built? I can't see it as part of the CI that runs for the kernel itself, e.g https://github.com/seL4/seL4/actions/runs/7380751924/job/20078457216.

Ivan-Velickovic:
Nevermind, my browser did not load the log file properly. I can see RPi3 being built for 32-bit and RPi4 being built for 64-bit. I think when I created the issue I meant to say that since we have both of those platforms we should also do run-time tests, but that can be a separate issue.

Basically just wanted a list of everything that either needs to be tested at run-time as well if we have the hardware, or only build-time if we don't.

PR #328, opened 2024-02-01
platforms: add polarfire hardware build

lsf37:
Ideally we'd also add simulation support, which first needs updates in the `simulate.py` script in [seL4_tools](https://github.com/seL4/seL4_tools). qemu seems to have support for it: <https://www.qemu.org/docs/master/system/riscv/microchip-icicle-kit.html>

I had a first go at that, but didn't get any output from qemu, so I'm leaving it out for now.

Adds one of the platforms listed in #282

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -67,6 +67,14 @@ platforms:
     req: [hifive] # , hifive1]
     march: rv64imac
 
+  POLARFIRE:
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
https://github.com/seL4/seL4/issues/1094 probably needs to be sorted out first before it will boot.

Ivan-Velickovic (approved):

PR #330, opened 2024-02-01
add ARMVIRT32 platform

axel-h:
This PR adds a AARCH32 QEMU platform. I am not sure how useful this is practically, given we have ARMv8 now and there might be less and less interest in ARMv7 support. We are testing the kernel on the QEMU/sabre simulation and the VM examples seem broken/neglected (https://github.com/seL4/camkes-vm-examples/issues/73). So there is not much use, except for some experimenting. And ensuring CI can hande the QEMU flexibility, as we need this for RISC-V then also.

### DIFF
--- a/camkes-vm/builds.yml
+++ b/camkes-vm/builds.yml
@@ -62,9 +62,9 @@ builds:
     success: "@xilinx-zcu102"
     settings:
         VmZynqmpPetalinuxVersion: '2022_1'
-- vm_minimal_ARMVIRT:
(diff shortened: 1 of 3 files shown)


axel-h:
Looking at https://github.com/axel-h/seL4/actions/runs/7783498425/job/21222088357?pr=140, this will work fine once https://github.com/seL4/seL4/pull/1182 is merged.

lsf37 (approved):

PR #332, opened 2024-02-10
factor out common workflows

lsf37:
With [reusable GitHub workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) we can now factor out common workflow files and maintain them only once here in the `ci-actions` repo.

Should also make it easier to deploy additional PR label actions to many/all of the seL4 org repos.

Unfortunately they *have* to live in the somewhat hidden `.github/workflows` directory, which I find a bit awkward, but it's still better than having to replicate everything. 

I have so far only tested the workflows that don't need secrets and trigger on `pull_request_target`, but those work. The display name changes, which mean we will have to update the names of required checks in all repos, but the benefit shoudl be worth the one-time effort.


lsf37:
Have now also tested `pull_request_target`, and it works. The only thing that is missing is testing secrets which only really works if the caller workflow has been merged.

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,116 @@
...
+# eventually.
+# Note that we could do the concurrency handling at a finer level, and only wrap
+# the actual run on the hardware. But there seems not much gain in letting the
+# older builds run, as these are usually obsolete with new pushes also.
+concurrency:
+  group: ${{ github.workflow }}-sel4test-pr-${{ github.event.number }}
+  cancel-in-progress: true
+
+jobs:
+  hw-build:
```
axel-h, line 37: Why does bench habe "freeze" and this one not?

.github/workflows/pr.yml
```diff
@@ -26,3 +26,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: seL4/ci-actions/bashisms@master
+
+  style:
+    name: Style
+    runs-on: ubuntu-latest
+    steps:
+    - uses: seL4/ci-actions/style@master
```
axel-h, line 34: Could this use local paths now?

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,116 @@
...
+    name: HW Build
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'push' ||
```
axel-h, line 40: We should remove the support for "push", as this workflow is not intended to be call on this event.

.github/workflows/pr.yml
axel-h: Seems this does not work as relative paths are resolved against the caller's repo. :(

.github/workflows/sel4bench-pr.yml
axel-h, line 1: Let's call this 'sel4bench-hw.yml` then to align the naming. I expect more "hw" and "sim" here for the other projects. Maybe hw/sim could become a "mode" parameter to project workflows?

.github/workflows/sel4test-hw.yml
lsf37: Probably copied from different sources. They should both have the "freeze" step so that we can be sure that all sub-tests are working on the same revisions.

.github/workflows/pr.yml
lsf37: yes, correct

.github/workflows/sel4bench-pr.yml
lsf37: I don't think sim/hw should be a parameter, because they have very different requirements and are fairly different workflows. Also simulation workflows, if they are not too many, could run by default without label, and the hardware ones should always be guarded on pull requests.

lsf37: For `sel4bench-hw` name: I'm not sure, actually. I also first changed it to `sel4bench-hw`, but then changed it back, because all `sel4bench` runs are on hardware, and the distinguishing factor for this workflow (as opposed to other sel4bench hardware workflows like the deployment one) is that it runs from pull requests.

Maybe the name consistency in the calling context is better than this reason, but `sel4bench-hw` doesn't really make much sense as a name in itself.

lsf37: let's just do the consistency thing and see how it goes, we can always rename it back.

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+      with:
+        manifest_repo: sel4test-manifest
+        manifest: master.xml
+        sha: ${{ github.event.pull_request.head.sha }}
+
+  hw-build:
+    name: HW Build
+    runs-on: ubuntu-latest
+    needs: code
+    # To reduce the load (especiually on the machine queue) we cancel any older
```
axel-h, line 49:
```suggestion
    # To reduce the load (especially on the machine queue) we cancel any older
```

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+        name: images-${{ matrix.march }}
+        path: '*-images.tar.gz'
+
+  hw-run:
+    name: HW Benchmark
+    if: ${{ github.repository_owner == 'seL4' }}
+    runs-on: ubuntu-latest
+    needs: [build]
+    concurrency:
+      group: ${{ github.workflow }}-sel4bench-pr-hw-run-${{ github.event.number }}
```
axel-h, line 71: We have to add `strategy.job-index` here again, otherwise the matrix runs kill their silibings.
```suggestion
      group: ${{ github.workflow }}-sel4bench-hw-run-pr-${{ github.event.number }}-${{ strategy.job-index }}
```

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+        sha: ${{ github.event.pull_request.head.sha }}
+
+  build:
+    name: Build
+    needs: code
+    runs-on: ubuntu-latest
+    # To reduce the load on GitHub runner numbers and machine queue we cancel
+    # any older runs of this workflow for the current PR.
+    concurrency:
+      group: ${{ github.workflow }}-sel4bench-pr-build-${{ github.event.number }}
```
axel-h, line 47: same here:
```suggestion
      group: ${{ github.workflow }}-sel4bench-build-pr-${{ github.event.number }}}-${{ strategy.job-index }}
```

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+    # runs of this workflow for the current PR.
+    concurrency:
+      group: ${{ github.workflow }}-sel4test-pr-build-${{ github.event.number }}
```
axel-h, line 52: same here:
```suggestion
      group: ${{ github.workflow }}-sel4test-build-pr-${{ github.event.number }}}-${{ strategy.job-index }}
```

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+    if: ${{ github.repository_owner == 'seL4' &&
+            (github.event_name == 'push' ||
+             github.event_name == 'pull_request_target' &&
+               github.event.action != 'labeled' &&
+               contains(github.event.pull_request.labels.*.name, 'hw-test') ||
+             github.event_name == 'pull_request_target' &&
+               github.event.action == 'labeled' &&
+               github.event.label.name == 'hw-test') }}
+    concurrency:
+      group: ${{ github.workflow }}-sel4test-pr-hw-run-${{ github.event.number }}
```
axel-h, line 101: same here:
```suggestion
      group: ${{ github.workflow }}-sel4test-hw-run-pr-${{ github.event.number }}-${{ strategy.job-index }}
```

.github/workflows/sel4test-sim.yml
```diff
@@ -0,0 +1,33 @@
...
+name: seL4Test-Sim
+
+on:
+  workflow_call:
+
+# To reduce the load we cancel any older runs of this workflow for the current
+# PR. Ok on the workflow level here, because this workflow does not re-trigger
+# on "labeled" events.
+concurrency:
+  group: ${{ github.workflow }}-sel4sim-pr-${{ github.event.number }}
```
axel-h, line 18: Debugging gives me `seL4Test-sel4sim-pr-` here when a never job cancels an old run. Debugging says
```
  "event_name": "pull_request",
    "event": {
      "action": "synchronize",
      ...
      "number": 140,      
      ...
```
so I'm a but puzzled why the PR number is missing there.

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+      matrix: ${{ steps.matrix.outputs.matrix }}
+    steps:
+    - id: matrix
+      uses: seL4/ci-actions/sel4test-hw-matrix@master
+
+  hw-run:
+    name: HW Run
+    runs-on: ubuntu-latest
+    needs: the_matrix
+    if: ${{ github.repository_owner == 'seL4' &&
```
axel-h, line 92: Since we have filtered almost everything in the `code` job, checking this seems enough:
```suggestion
    if: ${{ github.repository_owner == 'seL4' &&
            contains(github.event.pull_request.labels.*.name, 'hw-test')  }}
```

This is the context for a "labeld" event. the new label is part of the array then already.
```
"github": {
  "event_name": "pull_request_target",
  "event": {
    "action": "labeled",
    "label": {
      ...
      "name": "hw-test",
      ...
    },
    "number": 140,
    "pull_request": {
      ...
      "labels": [
        {
          ...
          "name": "hw-test",
          ...          
        }
      ],
...
```

.github/workflows/sel4test-sim.yml
axel-h: Seems the whole `github.event` does not exist a this stage.

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+
+jobs:
+  code:
+    name: Freeze Code
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request_target' &&
+              github.event.action != 'labeled' &&
+              (contains(github.event.pull_request.labels.*.name, 'hw-build') ||
+               contains(github.event.pull_request.labels.*.name, 'hw-test')) ||
+            github.event_name == 'pull_request_target' &&
```
axel-h, line 31: Isn't this redundant?

lsf37: if the workflow is used correctly probably yes, but if we use a wrong trigger then the rest of the event context is different and not safe to access, so I though it better to make sure it does not trigger at all instead of triggering based on data that might or might not make sense.

.github/workflows/sel4test-hw.yml
lsf37: Are you sure the label wasn't in there before from being set previously? Because I remember doing exactly this debug print and had to specifically distinguish the two cases because in one of them it wasn't in the list.

Maybe they changed this in the meantime to simplify this case. It'd be nice to know for sure what we can rely on. Will have a look at what the docs say.

lsf37: The [docs say](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=labeled#pull_request) that a labeled event must have a `pull_request` field which must have a `labels` field. I guess the only remaining question is if we are guaranteed that the new label is part of that `labels` field. If yes, we really can simplify.

axel-h: > Are you sure the label wasn't in there before from being set previously? 

Yes, I have trief this a couple of times as I was wondering exactly about that. Not sure if there is a guarantee, tough.

.github/workflows/sel4bench-hw.yml
axel-h: Actually, even this might not be enough if we are planning to have multiple repos calling this, where the workflow has the same name. We have to even add the `<org>/<repo>` here also to avoid the corner killing case where this is triggered by two PRs from different repos which happen to have the same ID.

.github/workflows/sel4test-hw.yml
axel-h: I meat grouping the labeld/not-labeld closer together using this:
```
github.event_name == 'pull_request_target' &&
((github.event.action != 'labeled' &&
  (contains(github.event.pull_request.labels.*.name, 'hw-build') ||
   contains(github.event.pull_request.labels.*.name, 'hw-test'))) ||
 (github.event.action == 'labeled' &&
  (github.event.label.name == 'hw-build' ||
   github.event.label.name == 'hw-test')))

```

.github/workflows/sel4test-hw.yml
lsf37: I figured out why it is they way it is: if the event is `labeled` the user might be adding a label that is a different label, e.g. `docs`. In that case we don't want to re-trigger the test if it is already running (and the label `hw-test` would already be in the list of labels, because it's already been set before).

.github/workflows/sel4bench-hw.yml
lsf37: Concurrency groups have [`repository` scope](https://github.com/orgs/community/discussions/78332), so we don't need to add that one at least. But we do need the job-index as you have it above.

.github/workflows/sel4test-hw.yml
lsf37: Yes, that would be fine. Maybe slightly easier to read.

.github/workflows/sel4test-sim.yml
lsf37: In which scenario is that happening? I.e. are you seeing this in the old job that is being cancelled or in the new job that is the reason for cancellation?

.github/workflows/sel4test-hw.yml
axel-h: True,  but this is properly filtered-out by the job `code` already - and I want to keep this part. So, when we arrive here in the job `hw-run`, it can only have happened due to a push or a new "hw_build"/"hw_test" label. So we can actually simply check if we have the "hw-test" set here.

.github/workflows/sel4bench-hw.yml
axel-h: Right, and I forgot we are executing in the caller's repo context here (and that's why relative paths here don't work either....)

.github/workflows/sel4test-sim.yml
axel-h: I'm seeing this in the old job, where github gives the cancellation reason. This is the only place where the group name is ever printed. It seems impossible to see the concurrency group for a running job anywhere. You can print it in a job/step, but the context there seems slightly different, because this runs on the runner. The workflow control and concurrency seems to run on an "action server", which does not see some dynamic parts of the contexts. The github documentation is quite thin here, but that is what I found in some github community postings.

.github/workflows/sel4test-hw.yml
lsf37: I don't think that's good. Why make assumptions if we can just test what we need to test? If we ever refactor or move this it'll work fine the way it is now, but not if we depend on context info from other jobs.

.github/workflows/sel4test-sim.yml
lsf37: Ok, the docs do say that `github`, `inputs`, and `vars` are available at the [workflow level](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#concurrency), and additionally `needs`, `strategy`, and `matrix` on the [job level](https://docs.github.com/en/actions/using-jobs/using-concurrency#using-concurrency-in-different-scenarios). So I think we are fine with this expression, it's just hard to debug as usual on GitHub.

.github/workflows/sel4test-hw.yml
axel-h: Well, there are pros and cons. We depend on the build step anyway at the moment. In a bigger picture, it would be nicer if a caller would decide when to invoke which step, but I have not found nice was for this. So let's keep it as it is for now.

.github/workflows/sel4test-sim.yml
axel-h: Let's see if I get some answers here: https://github.com/orgs/community/discussions/107552

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+        sha: ${{ github.event.pull_request.head.sha }}
+
+  build:
+    name: Build
+    needs: code
+    runs-on: ubuntu-latest
+    # To reduce the load on GitHub runner numbers and machine queue we cancel
+    # any older runs of this workflow for the current PR.
+    concurrency:
+      group: ${{ github.workflow }}-sel4bench-build-pr-${{ github.event.number }}}-${{ strategy.job-index }}
```
axel-h, line 47: one `}` too much
```suggestion
      group: ${{ github.workflow }}-sel4bench-build-pr-${{ github.event.number }}-${{ strategy.job-index }}
```

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,124 @@
...
+        sha: ${{ github.event.pull_request.head.sha }}
+
+  hw-build:
+    name: HW Build
+    runs-on: ubuntu-latest
+    needs: code
+    # To reduce the load (especially on the machine queue) we cancel any older
+    # runs of this workflow for the current PR.
+    concurrency:
+      group: ${{ github.workflow }}-sel4test-build-pr-${{ github.event.number }}}-${{ strategy.job-index }}
```
axel-h, line 52: one `}` too much
```suggestion
      group: ${{ github.workflow }}-sel4test-build-pr-${{ github.event.number }}}-${{ strategy.job-index }}
```

.github/workflows/sel4test-sim.yml
lsf37: I'm going to put it into the job level for now, so that we can merge this. Would still be good to figure out what exactly is available, though.

lsf37:
I think this is ready to merge now. We can move the concurrency group for `sel4test-simp` back up to the workflow level when we get a definitive answer in <https://github.com/orgs/community/discussions/107552>.

.github/workflows/sel4test-sim.yml
axel-h: The other option could be adding another callable workflow layer here. The the job level concurrency group there will cover everything. It could also filter the events and invoke this workflow with a simple bool parameter for enabling hw-test.

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+
+# downgrade permissions to read-only as you would have in a standard PR action
+permissions:
+  contents: read
+
+jobs:
+  code:
+    name: Freeze Code
+    runs-on: ubuntu-latest
+    if: ${{ github.event_name == 'pull_request_target' &&
```
axel-h, line 24: Apply the same as in `sel4test-hw.yml` here also:
```
    if: ${{ github.event_name == 'pull_request_target' &&
            (
              github.event.action != 'labeled' &&
              (contains(github.event.pull_request.labels.*.name, 'hw-build') ||
               contains(github.event.pull_request.labels.*.name, 'hw-test')) ||
              github.event.action == 'labeled' &&
              (github.event.label.name == 'hw-build' ||
               github.event.label.name == 'hw-test')
            ) }}
```

.github/workflows/sel4test-sim.yml
```diff
@@ -0,0 +1,29 @@
...
+#
+# See sel4test-sim/builds.yml in the repo seL4/ci-actions for configs.
+
+name: seL4Test-Sim
+
+on:
+  workflow_call:
+
+jobs:
+  sim:
```
axel-h, line 15: It seem the "test with" in PRs works well for "seL4test", "seL4bench" and "sim". But I'm not fully understanding when we have to do a code freeze step an when not. Seems we have 3 different way how things are done for seL4test, seL4bench and sim. So a follow up task once this is merged should be aligning this better.

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,125 @@
...
+      with:
+        name: images-${{ matrix.march }}-${{ matrix.compiler }}
+        path: '*-images.tar.gz'
+    - name: Upload kernel.elf files
+      uses: actions/upload-artifact@v4
+      with:
+        name: kernel.elf-${{ matrix.march }}-${{ matrix.compiler }}
+        path: '*-kernel.elf'
+
+  the_matrix:
```
axel-h, line 79: We could make this a step in the `code` job also. It avoids spawning another runner VM.
```
jobs:
  code:
    ...
    outputs:
      ...
      matrix: ${{ steps.matrix.outputs.matrix }}
    steps:
    ...
    - id: matrix
      uses: seL4/ci-actions/sel4test-hw-matrix@master

    ...
  hw-run:
    ...
    needs: hw-build
```

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,125 @@
...
+    - name: Build
+      uses: seL4/ci-actions/sel4test-hw@master
+      with:
+        xml: ${{ needs.code.outputs.xml }}
+        march: ${{ matrix.march }}
+        compiler: ${{ matrix.compiler }}
+    - name: Upload images
+      uses: actions/upload-artifact@v4
+      with:
+        name: images-${{ matrix.march }}-${{ matrix.compiler }}
```
axel-h, line 71:
```suggestion
        name: sel4test-images-${{ matrix.march }}-${{ matrix.compiler }}
```
I have done some experimenting with running both `sel4test` and `sel4bench`. Since they are all part of the same workflow now, there can be upload conflicts here. 
- `sel4bench` uploads as `images-${{ matrix.march }}`
- `sel4test` uploads as `images-${{ matrix.march }}-${{ matrix.compiler }}`

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,125 @@
...
+    steps:
+      - name: Get machine queue
+        uses: actions/checkout@v4
+        with:
+          repository: seL4/machine_queue
+          path: machine_queue
+      - name: Download image
+        uses: actions/download-artifact@v4
+        with:
+          name: images-${{ matrix.march }}-${{ matrix.compiler }}
```
axel-h, line 116:
```suggestion
          name: images-sel4test-${{ matrix.march }}-${{ matrix.compiler }}
```

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+    steps:
+    - name: Build
+      uses: seL4/ci-actions/sel4bench@master
+      with:
+        xml: ${{ needs.code.outputs.xml }}
+        march: ${{ matrix.march }}
+    - name: Upload images
+      uses: actions/upload-artifact@v4
+      with:
+        name: images-${{ matrix.march }}
```
axel-h, line 62:
```suggestion
        name: images-sel4bench-${{ matrix.march }}
```

.github/workflows/sel4bench-hw.yml
```diff
@@ -0,0 +1,118 @@
...
+          path: machine_queue
+      - name: Get march
+        id: plat
+        uses: seL4/ci-actions/march-of-platform@master
+        with:
+          platform: ${{ matrix.platform }}
+      - name: Download image
+        uses: actions/download-artifact@v4
+        with:
+          name: images-${{ steps.plat.outputs.march }}
```
axel-h, line 104:
```suggestion
          name: images-sel4bench-${{ steps.plat.outputs.march }}
```

.github/workflows/sel4test-hw.yml
```diff
@@ -0,0 +1,125 @@
...
+        compiler: ${{ matrix.compiler }}
+    - name: Upload images
+      uses: actions/upload-artifact@v4
+      with:
+        name: images-${{ matrix.march }}-${{ matrix.compiler }}
+        path: '*-images.tar.gz'
+    - name: Upload kernel.elf files
+      uses: actions/upload-artifact@v4
+      with:
+        name: kernel.elf-${{ matrix.march }}-${{ matrix.compiler }}
```
axel-h, line 76:
```suggestion
        name: kernel.elf-sel4test-${{ matrix.march }}-${{ matrix.compiler }}
```

.github/workflows/sel4test-hw.yml
lsf37: No, I don't think that's a good idea, those names are partially used in other scripts that people have and are used to. I don't see any benefit in changing them, they are local to their specific workflow and make sense in their context.

.github/workflows/sel4test-hw.yml
lsf37: No. This is supposed to be factoring out existing workflows not a complete redesign of everything. Keeping this step separate keeps those concerns separated. The code job should get the code, not the matrix.

.github/workflows/sel4test-sim.yml
lsf37: `Test with` works with all of them. It's part of the checkout scripts they all call. Code freeze just eliminates a race condition where one test job might see a different manifest to another one (so we get the one manifest in the beginning and give it to all jobs).

.github/workflows/sel4test-hw.yml
axel-h: We have only one shared worflow `sel4test-hw.yml` in the sel4 kernel project now that triggers both "sel4test" and "sel4bench" . So all these artifacts end up together, see e.g. the artifact list. At the moment with just these it works because the naming pattern is different. If another project is added here, it need to use a better pattern. And if we consider running "sel4bench" for the compilers it will also clash badly. Luckly the upload action warns on duplicate names now instead of silently overwriting them.

lsf37: The upload action actually fails now if there is a duplicate. It's true that they will end up in the same summary screen. We'd still need to track down what depends on this, so we should not just change artifact names if we don't need to. If it does become a problem in the future, we can still change things then (and it's then only one central workflow, so it's much easier to do).

We don't need to get everything perfect now, we won't achieve that anyway. We should observe this in action for a while and then see what we need.

lsf37: Ok, I can see how the UI will be confusing for the `images` part. I don't think those are used in automation outside this workflow (but kernel.elf might be). I'll try to double check on that first part and change those.

axel-h (approved):
It's good to merge this in the current state. I've done quite a lot of experimenting now and in general this works. So no gain in blocking the merge.

PR #337, opened 2024-02-27
builds.py: let mq lock be reclaimed after 30min

lsf37:
A typical job takes about 5-10 min. If the lock has been held for 30 min in a single job, we assume something has gone wrong and the lock should be released so other jobs (or humans needing the machine) can proceed.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -451,8 +451,8 @@ def mq_run(success_str: str,
 
 
 def mq_lock(machine: str) -> List[str]:
-    """Get lock for a machine."""
-    return ['time', 'mq.sh', 'sem', '-wait', machine, '-k', job_key()]
+    """Get lock for a machine. Allow lock to be reclaimed after 30min."""
+    return ['time', 'mq.sh', 'sem', '-wait', machine, '-k', job_key(), '-T', '1800']
 
 
 def mq_release(machine: str) -> List[str]:


seL4-platforms/builds.py
```diff
@@ -451,8 +451,8 @@ def mq_run(success_str: str,
 
 
 def mq_lock(machine: str) -> List[str]:
-    """Get lock for a machine."""
-    return ['time', 'mq.sh', 'sem', '-wait', machine, '-k', job_key()]
+    """Get lock for a machine. Allow lock to be reclaimed after 30min."""
+    return ['time', 'mq.sh', 'sem', '-wait', machine, '-k', job_key(), '-T', '1200']
```
axel-h, line 455: Is this seconds, so 1800 then for 30 mins?

axel-h:
I wonder we shoud combine this with also using https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes

lsf37:
> I wonder we shoud combine this with also using https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepstimeout-minutes

That's a different use case, and a bit harder to give a good value for, because that workflow timeout is the overall time for all jobs -- that is much more variable than a single job once it got the lock, because the global one also includes lock wait time and varies depending on machine queue contention. The time for a single job once it has the lock is reasonably predictable. 

Might still be worth looking into and reducing the default for some of them.

The main case that this specific timeout is a fix for is the condition where a job takes the lock, and then the machine queue server becomes unavailable on the network when it's trying to release it. AFAICS there is no fix for that apart from a timeout. It doesn't happen that often, but it's still happened maybe 4 times or so in the last two years (mostly around network outages, unsurprisingly).

axel-h (approved):
Look good to me.

kent-mcleod:
We used to handle this by setting the timeout on the job that was running.  The step to release the locks would trigger when the job completed or was cancelled or was timed out.  Would that not work for us? Is releasing the locks triggered as a :post action? Do post actions run when earlier steps timeout?

lsf37:
> We used to handle this by setting the timeout on the job that was running. The step to release the locks would trigger when the job completed or was cancelled or was timed out. Would that not work for us? Is releasing the locks triggered as a :post action? Do post actions run when earlier steps timeout?

They do, and all of this is already done in the code. This if for when that has failed, because the network is unavailable, so the normal lock release fails and then the post step clear-all lock release fails, and the lock is still held on the server when it comes back on the network.

seL4-platforms/builds.py
lsf37: Indeed, that doesn't match. I'll change it to 1800 seconds, because that's fewer places to change and this should only be necessary in very rare cases anyway. If jobs fail or get stuck in a "normal" way (i.e. no network outage) there are already the standard timeouts in place for that.

lsf37:
> Either change the title to '20 minutes' or change the timeout to 1800 seconds. Either would be enough for this to pass.

Done. I've also updated the commit message to explain that this is a timeout of last resort.

wom-bat (approved):
Looks good now

PR #340, opened 2024-03-01
simulation: signal failure on timeout

lsf37:
Return exit code 0 on expected success string, and exit code 1 on timeout for all simulation scripts. Reduce standard timeout to 20min.

Previously, failure to produce the success string would lead to a timeout which did not signal a test failure and would just continue on to the next test.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -27,7 +27,8 @@ import sys
 
 # exported names:
 __all__ = [
-    "Build", "load_builds", "run_builds", "run_build_script", "junit_results", "sanitise_junit"
+    "Build", "load_builds", "run_builds", "run_build_script", "junit_results", "sanitise_junit",
+    "sim_script"
 ]
 
 # where to expect jUnit results by default
@@ -550,6 +551,14 @@ parsed_junit_results = 'parsed_results.xml'
 # expects to run in build dir of a standard seL4 project setup
 sanitise_junit = ["python3", "../projects/seL4_libs/libsel4test/tools/extract_results.py",
                   "-q", junit_results, parsed_junit_results]
+
+
+def sim_script(success: str, timeout=1200):
+    return [
+        "expect", "-c",
+        'spawn ./simulate; set timeout %d; expect { "%s" {exit 0} timeout {exit 1} }' %
+        (timeout, success)
+    ]
 
 
 def run_build_script(manifest_dir: str,
--- a/sel4test-sim/build.py
+++ b/sel4test-sim/build.py
@@ -18,13 +18,13 @@ import sys
 def run_simulation(manifest_dir: str, build: Build):
     """Run one simulation build and test."""
 
-    expect = f"\"{build.success}\""
+    expect = '{ "%s" {exit 0} timeout {exit 1} }' % build.success
 
     script = [
         ["../init-build.sh"] + build.settings_args(),
         ["ninja"],
         ["bash", "-c",
-         f"expect -c 'spawn ./simulate; set timeout 3000; expect {expect}' | tee {junit_results}"]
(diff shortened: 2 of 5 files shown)


sel4test-sim/build.py
```diff
@@ -18,13 +18,13 @@
...
 
-    expect = f"\"{build.success}\""
+    expect = '{ "%s" {exit 0} timeout {exit 1} }' % build.success
 
     script = [
         ["../init-build.sh"] + build.settings_args(),
         ["ninja"],
         ["bash", "-c",
-         f"expect -c 'spawn ./simulate; set timeout 3000; expect {expect}' | tee {junit_results}"]
+         f"expect -c 'spawn ./simulate; set timeout 1200; expect {expect}' | tee {junit_results}"]
```
axel-h, line 27: Nitpicking: could re-use things here also?
```suggestion
         f"{' '.join(sim_script(build_success))} | tee {junit_results}"
```

lsf37: Also just joining the parts of the list with space is not guaranteed to result in a valid command line. You'd need to quote each part separate and make sure that the quote character used is not used on the inside of the argument. I wouldn't go there.

axel-h: An no, there is another quote layer in there. Maybe a long term improvement could be using https://pexpect.readthedocs.io/en/stable/install.html

seL4-platforms/builds.py
```diff
@@ -552,6 +553,14 @@ def summarise_junit(file_path: str) -> Tuple[int, List[str]]:
                   "-q", junit_results, parsed_junit_results]
 
 
+def sim_script(success: str, timeout=1200):
+    return [
+        "expect", "-c",
+        'spawn ./simulate; set timeout %d; expect { "%s" {exit 0} timeout {exit 1} }' %
```
axel-h, line 559: Nitpicking: could break the expect tcl commands down into a list:
```suggestion
        "; ".join([
             "spawn ./simulate",
             f"set timeout {timeout}",
             'expect { "%s" {exit 0} timeout {exit 1} }' % success'])
```

lsf37: I don't think that's clearer, and it's dangerous if later somebody wants to add something the is not `;` terminated and doesn't spot what exactly the `join` uses.

The commands in a script are a list, because the command line invocation at the consuming end expects a list, not because a list was nicer to write.

axel-h: I can see you point. Maybe give https://pexpect.readthedocs.io/en/stable/install.html a try to make this nice then.

lsf37: Using `pexpect` doesn't fit into the framework, instead of executing a list of external commands it would now sometimes need to do a python thing in between. That sounds very messy to me.

I find the current version perfectly fine and readable, to be honest. The problem in the previous script wasn't that it wasn't readable, it was that whoever wrote the original version 10 years ago didn't read the manual for `expect` properly.

axel-h: I'm biassed there because we did create a python framework around running simulation on QEMU and other simulators. At some point expect was not powerful enough and we needed a more interactive way.

lsf37: For a script specifically for simulators that makes a lot of sense. When we want to go there, I'd put that into a separate program that is then called from this framework, though. I don't think it should be part of the general CI harness, because then you also can't run it yourself on the command line.

lsf37: (Currently we're not planning to add that kind of complexity to the simulation runs, though)

axel-h (approved):
This looks good. I am doing a few more test currently, but it seems github has some general CI issues that prevent actions from running...

wom-bat (approved):
Looks good to me.
At some point we may want to switch to the Python Pexpect module instead of relying on an external tool.

Issue #339, opened 2023-04-16
vm_minimal_sim run failure is not reported


axel-h:
The build of `vm_minimal_sim` also runs the QEMU simulation as part opf the build job. Unfortunately, failures there are not reported. While the script clearly expects `buildroot login:` and this never comes, it seems this does not return an error? A quick look at the framkework seems to handle errors:
- https://github.com/seL4/ci-actions/blob/master/camkes-vm/build.py#L42-L45
- `run_build_script()` at https://github.com/seL4/ci-actions/blob/master/seL4-platforms/builds.py#L609-L612

For a run that should have see for example https://github.com/seL4/camkes-vm/actions/runs/4714879145/jobs/8361496520?pr=81
```
-----------[ start test vm_minimal_sim ]-----------
  +++ ../init-build.sh -DAARCH64=TRUE -DPLATFORM=qemu-arm-virt -DCAMKES_VM_APP=vm_minimal
...
  -- Generating done
  -- Build files have been written to: /github/workspace/build
  +++ ninja
  [1/407] Generating capdl/capdl-loader-app/gen_config/capdl_loader_app/platform_info.h
...
  [407/407] Generating images/capdl-loader-image-arm-qemu-arm-virt
  +++ tar czf ../vm_minimal_sim-images.tar.gz images/
  +++ bash -c expect -c 'spawn ./simulate; set timeout [3000](https://github.com/seL4/camkes-vm/actions/runs/4714879145/jobs/8361496520?pr=81#step:3:3004); expect "buildroot login:"'
  spawn ./simulate
  ./simulate: qemu-system-aarch64 -machine virt,virtualization=on,highmem=off,secure=off -cpu cortex-a53 -nographic  -m size=2048  -kernel images/capdl-loader-image-arm-qemu-arm-virt
  ELF-loader started on CPU: ARM Ltd. Cortex-A53 r0p4
...  
  Bootstrapping kernel
...  
  Booting all finished, dropped to user space
...  
  FAULT HANDLER: data fault from vm0.vm0_0_control (ID 0x2) on address 0x90, pc = 0x464014, fsr = 0x93800006
...
  -----------[ end test vm_minimal_sim ]-----------
vm_minimal_sim succeeded
```

wom-bat:
The `simulate` scriupt is doing what it's meant to; but the test needs to look at the output and check it's getting to guetst userspace

lsf37:
[...]

PR #341, opened 2024-03-14
Sim scripts: Fix issue 340

wom-bat:
Add success criteria for each simulation run.  Otherwise the string is empty and never matches.

### DIFF
--- a/camkes-vm/build.py
+++ b/camkes-vm/build.py
@@ -8,7 +8,7 @@ Parse builds.yml and run CAmkES VM test on each of the build definitions.
 Expects seL4-platforms/ to be co-located or otherwise in the PYTHONPATH.
 """
 
-from builds import Build, run_build_script, run_builds, load_builds, release_mq_locks, SKIP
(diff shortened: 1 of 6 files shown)


axel-h:
see 
- https://github.com/seL4/ci-actions/issues/339 
- https://github.com/seL4/ci-actions/pull/340

lsf37:
That can't be the reason for sel4test, where it is provided, but still failing.

lsf37:
Still need to confirm with a debug run, but the success string is provided [here](https://github.com/seL4/ci-actions/blob/49c59bf2b22ff88fecf7ba8da61404bbef6d620a/camkes-vm/builds.yml#L7-L9) as default:

```yml
default:
  success: 'buildroot login:'
  error: 'Halting...'
```

(which reminds me that we should also match for the failure case and terminate early there instead of timing out -- Bamboo didn't do that, which is why it's not in the current scripts, but it should be easy to add once we've figured out what is going on)

lsf37:
Closing this one, since it's not changing behaviour on the test.

PR #351, opened 2024-04-11
python: remove broken unused function

axel-h:
Commit 85c8f71d should never have added this function, it slipped the review, The function's type annotation is broken, it returns a dict and not a string. Furthermore, there is not need for this wrapper, as we can (and do) read the attribute directly. Nobody uses this function, so let's just remove it.

### DIFF
--- a/seL4-platforms/platforms.py
+++ b/seL4-platforms/platforms.py
@@ -160,9 +160,6 @@ class Platform:
     def cmake_toolchain_setting(self, mode: int) -> str:
         return self.toolchain_arch_str() + str(mode)
 
-    def settings(self) -> str:
-        return self.settings
-
     def get_image_platform(self, mode: int) -> str:
         return self.image_platform or self.get_platform(mode)
 


lsf37:
The type annotation might be wrong, but it did not slip review -- we left it in because we were planning to use it in the future.

lsf37:
I do agree that it's not ideal. If we want it to be an interface function, it should probably be `get_settings`. Ok, so I'm fine with removing it for now (leaving in the rest of https://github.com/seL4/ci-actions/commit/85c8f71d0dd62f4b7861aacd0bd19056799a9cb6) until we need an actual interface function, and then make sure that that function is used everywhere.

lsf37 (approved):

axel-h:
I am trying to make sense if the current scripts and it seems quit confusing sometimes. Some scripts seems to need a update to be aligned on the latest features. Ans there seems some legacy where cleaning this up might simplify things a lot. There will be a few more smaller PR. These might conflict with what you have in mind, but I'd appreciate feedback explaining where you are heading. At the moment some abstraction layers just seem to make things overly complicated.

lsf37:
I'm not really a fan of fixing things that are not broken to understand them. Do you have any specific questions?

axel-h:
I think its best to discuss this in the PRs. Things are not broken, but a bit confusing what to do and why.

PR #357, opened 2024-07-18
allow other manifests in l4v-deploy

lsf37:
Add mechanism for deploying after preprocess and after proof test to manifests other than `devel.xml`/`default.xml`. Currently only `mcs-devel.xml`/`mcs.xml` are allowed.

### DIFF
--- a/l4v-deploy/common.py
+++ b/l4v-deploy/common.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright 2020, Data61, CSRIO
+# Copyright 2020, Data61, CSIRO
 # SPDX-License-Identifier: BSD-2-Clause
 #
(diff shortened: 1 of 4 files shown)


l4v-deploy/common.py
```diff
@@ -7,6 +7,7 @@
 
 import subprocess
 
+
```
corlewis, line 10: Accidental newline? The previous commit removes it.

corlewis (approved):

l4v-deploy/common.py
lsf37: It got added by the style check -- I shouldn't have removed it in the first commit.

lsf37: Will see if I can reshuffle that.

PR #358, opened 2024-07-26
sel4test-hw: temporarily pin docker base image

lsf37:
Base images since 2024-07-19 produce sel4test images with new failures for imx8mm-evk and odroid_c4. Pin to the version before that to enable continued development for now.

The intention is to revert this commit once we have figured out what has changed in the build container to make the tests fail.


axel-h (approved):
Go for it then.

Ivan-Velickovic (approved):

PR #359, opened 2024-08-08
zynqmp: enable hyp tests

lsf37:
Hypervisor features are listed as supported on the website for the `zynqmp` platforms and do seem to be working fine in manual tests.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -237,6 +237,7 @@ platforms:
     arch: arm
     modes: [64]
     smp: [64]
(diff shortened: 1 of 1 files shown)


Xaphiosis (approved):
That table on the website could use an update (the "ARM HYP" vs "Yes" ... no idea)

lsf37:
> That table on the website could use an update (the "ARM HYP" vs "Yes" ... no idea)

We'll get to that one in the website revamp in any case, so I think I'll leave that untouched for now.

PR #362, opened 2024-08-13
camkes-vm: avoid exception when BAMBOO is not set

lsf37:
Since we are enumerating builds, it is possible that BAMBOO is not set, leading to an exception when deleting this (unused) key. Therefore check for presence before deleting.

(#361 broke the camkes-vm build because of this)


axel-h (approved):
Are there other scripts here that also do this unconditionally, too?

lsf37:
> Are there other scripts here that also do this unconditionally, too?

At least none have failed yet, but you're right I should have a look at that.

PR #369, opened 2024-10-12
matrix: temporarily remove riscv/clang

lsf37:
Temporarily remove riscv/clang until clang-12 works.


alwin-joshy (approved):

lsf37:
Abandoning this one, because it seems the clang-12 excursion didn't work after all.

PR #372, opened 2024-12-02
platforms: add imx93

lsf37:
Currently there are no boards yet in CI for this platform. (To be changed in the next few months).

Should be merged after the following PRs are merged:

- https://github.com/seL4/seL4/pull/1356
- https://github.com/seL4/seL4_tools/pull/211
- https://github.com/seL4/util_libs/pull/189
- https://github.com/seL4/sel4bench/pull/50

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -157,6 +157,15 @@ platforms:
     req: [tqma]
     march: armv8a
 
+  IMX93:
+    arch: arm
+    modes: [64]
+    smp: [64]
+    aarch_hyp: [64]
+    platform: imx93
+    march: armv8a
+    no_hw_test: true
+
   OMAP3:
     arch: arm
     modes: [32]


Ivan-Velickovic:
Does it work with the problematic combinations like SMP + HYP + MCS?

I guess it doesn't matter until we actually get the hardware.

Ivan-Velickovic (approved):

lsf37:
> Does it work with the problematic combinations like SMP + HYP + MCS?
> 
> I guess it doesn't matter until we actually get the hardware.

Yup, it works on SMP + HYP + MCS (Indan does have the hardware, the plan is to donate it to the foundation when  the port is complete).

Indanz:
Yes, all combinations I tried work for me. Only configuration not supported is Aarch32.

lsf37:
Was there a blocker for AArch32 or would it just be additional work? (I don't think we need it, just curious)

Indanz:
I think the GICv3 code was only tested with 64-bits, but we could set it in backward compatibility mode and use it the GICv2 driver on Aarch32. Or just review and test the GICv3 kernel driver for 32-bit. So no real blockers, just some additional work one way or the other.

Ivan-Velickovic:
> I think the GICv3 code was only tested with 64-bits. But we could set the GIC in backward compatibility mode and use the GICv2 driver on Aarch32. Or just review and test the GICv3 kernel driver for 32-bit. So no real blockers, just some additional work one way or the other.

I thought we ran the i.MX8MM-EVK in 32-bit mode which has GICv3?

Ivan-Velickovic:
https://github.com/seL4/ci-actions/blob/965658befb14b9515979d65bec95337bbe7232b0/seL4-platforms/platforms.yml#L144-L150

Indanz:
I did a quick test with the below patch applied, but for me it hangs in Elfloader with no output.
The command I used to build was:

`$ ../test/init-build.sh -DPLATFORM=imx93 -DKernelArmExportPCNTUser=ON -DKernelArmExportPTMRUser=ON -DRELEASE=OFF -DMCS=ON -DAARCH32=TRUE -DCROSS_COMPILER_PREFIX=arm-linux-gnueabihf- ElfloaderArmV8LeaveAarch64=ON`

```diff
diff --git a/libsel4/sel4_plat_include/imx93/sel4/plat/api/constants.h b/libsel4/sel4_plat_include/imx93/sel4/plat/api/constants.h
index 0496273f6..c22c1eefd 100644
--- a/libsel4/sel4_plat_include/imx93/sel4/plat/api/constants.h
+++ b/libsel4/sel4_plat_include/imx93/sel4/plat/api/constants.h
@@ -7,3 +7,8 @@
 
 #include <sel4/config.h>
 #include <sel4/arch/constants_cortex_a55.h>
+
+#if CONFIG_WORD_SIZE == 32
+/* First address in the virtual address space that is not accessible to user level */
+#define seL4_UserTop 0xe0000000
+#endif
diff --git a/src/arch/arm/armv/armv8-a/32/machine_asm.S b/src/arch/arm/armv/armv8-a/32/machine_asm.S
index 671d0de81..eabf76ebf 100644
--- a/src/arch/arm/armv/armv8-a/32/machine_asm.S
+++ b/src/arch/arm/armv/armv8-a/32/machine_asm.S
@@ -7,9 +7,9 @@
 #include <config.h>
 #include <machine/assembler.h>
 
-#if defined(CONFIG_ARM_CORTEX_A53)
+#if defined(CONFIG_ARM_CORTEX_A53) || defined(CONFIG_ARM_CORTEX_A55)
 
-/* A53 hardware does not support TLB locking */
+/* ARMv8 hardware does not support TLB locking */
 BEGIN_FUNC(lockTLBEntry)
     bx lr
 END_FUNC(lockTLBEntry)
diff --git a/src/plat/imx93/config.cmake b/src/plat/imx93/config.cmake
index 5d6ed8faf..b405617d4 100644
--- a/src/plat/imx93/config.cmake
+++ b/src/plat/imx93/config.cmake
@@ -9,15 +9,21 @@ cmake_minimum_required(VERSION 3.7.2)
 declare_platform(imx93 KernelPlatformIMX93 PLAT_IMX93 KernelArchARM)
 
 if(KernelPlatformIMX93)
-    declare_seL4_arch(aarch64)
+    declare_seL4_arch(aarch64 aarch32 arm_hyp)
     set(KernelArmCortexA55 ON)
     set(KernelArchArmV8a ON)
     set(KernelArmGicV3 ON)
     config_set(KernelARMPlatform ARM_PLAT ${KernelPlatform})
     list(APPEND KernelDTSList "tools/dts/${KernelPlatform}.dts")
-    list(APPEND KernelDTSList "src/plat/imx93/overlay-${KernelPlatform}.dts")
+    if(KernelSel4ArchAarch32)
+        list(APPEND KernelDTSList "src/plat/imx93/overlay-${KernelPlatform}-32.dts")
+    else()
+        list(APPEND KernelDTSList "src/plat/imx93/overlay-${KernelPlatform}.dts")
+    endif()
     declare_default_headers(
         TIMER_FREQUENCY 24000000
+        CLK_MAGIC 2863311531llu
+        CLK_SHIFT 36u
         TIMER drivers/timer/arm_generic.h
         TIMER_OVERHEAD_TICKS 1
         NUM_PPI 32
```
The alternative DTS file is necessary to avoid overflowing 32-bits on the memory range.

Any suggestions?

Ivan-Velickovic:
@Indanz I don't fully understand why this is the case for the i.MX8 based boards in AArch32 mode but it might be worth a shot doing something like this:

https://github.com/seL4/seL4_tools/blob/bef85f32a4829c5062fd6c808c7cab57b816cd9f/cmake-tool/helpers/application_settings.cmake#L40-L48

Indanz:
> @Indanz I don't fully understand why this is the case for the i.MX8 based boards in AArch32 mode

The `ElfloaderArmV8LeaveAarch64` is to tell Elfloader that it will start in 64-bit mode and hence needs to exit to 32-bit mode before it can proceed.

The comment is talking about `ElfloaderImage` needing to be set to `binary`, instead of `uimage`. I've only tried `elf`, which didn't work for me with `bootelf`. Why that doesn't work may also apply to `uimage` I guess, but haven't found out why. I don't know if setting `IMAGE_START_ADDR` somehow forces the image to be binary automatically or something.

Anyway, thanks for the hint, I'll fiddle with `IMAGE_START_ADDR` and see if that helps.

Indanz:
When setting `-DElfloaderArmV8LeaveAarch64=ON` (with the `-D` part) and using `IMAGE_START_ADDR` I can boot, but I get a mysterious page fault in sel4test, which seems weird enough to spend more time figuring out what's going on. 

Basically the `seL4_ARM_Page_Map()` calls for virtual address 0xdf7fe000 succeeds, but the memset causes a VM fault. It's the same virtual address that faults, different physical addresses if I use different load addresses. I guess it's the first page that gets mapped, but haven't confirmed that yet. The fault status is "Translation fault, level 1", so I guess something goes wrong with setting up the TLBs (but then how did we got so far into user space boot?).

lsf37:
Will merge this now for 64 bit only, but happy to add 32 when that works.

Indanz:
> Will merge this now for 64 bit only, but happy to add 32 when that works.

Do you want me to put more time into that? Currently it's on hold. (And if yes, should I book the time on the imx93 project or the seL4 foundation?)

lsf37:
The project as such is finished, so it would have to be on foundation time. As a low priority task, I would say yes. It would be nice to have 32 bit mode working eventually, even if there is no big demand/hurry.

PR #378, opened 2025-02-18
js: update dependencies

lsf37:
Fixes https://github.com/seL4/ci-actions/security/dependabot/21

Dependency update only (via `npm audit --fix`)

### DIFF
--- a/js/node_modules/.package-lock.json
+++ b/js/node_modules/.package-lock.json
@@ -52,9 +52,9 @@
       }
     },
     "node_modules/undici": {
-      "version": "5.28.2",
-      "resolved": "https://registry.npmjs.org/undici/-/undici-5.28.2.tgz",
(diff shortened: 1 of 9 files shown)


Indanz (approved):

corlewis (approved):
Not related to this PR, but it does look to be a valid links failure. Guessing purely based on the files that exist there, but it maybe should be https://deb.debian.org/debian/pool/main/d/devscripts/devscripts_2.23.7~bpo11+1.tar.xz

lsf37:
> Not related to this PR, but it does look to be a valid links failure. Guessing purely based on the files that exist there, but it maybe should be https://deb.debian.org/debian/pool/main/d/devscripts/devscripts_2.23.7~bpo11+1.tar.xz

This one is fixed in #377

PR #379, opened 2025-02-26
sel4test: enable SMP tests

lsf37:
With some luck https://github.com/seL4/seL4/pull/871 fixes all of these.

Should only be merged after this is confirmed on https://github.com/seL4/seL4/pull/871


Indanz (approved):

lsf37:
It is traditional to push to production on a Fri evening, so here goes. 🚀 

Without this change, the tests are blocked on Odroid C4 now failing SCHED_CONTEXT_0014 on gcc, so merging this should not make anything worse..

PR #381, opened 2025-02-27
fix setuptools version

lsf37:
`sel4-deps` have started to fail installation on GitHub CI, although they seem to work fine on clean Ubuntu 22.04.

See also the saga on https://github.com/pypa/setuptools/issues/4483 -- fixing the version of `setuptools` seems to fix the problem for CI.


lsf37:
Have tested this with the style check on this PR (commit now dropped) and it passed. I have not consistently see it fail everywhere, though, so it's unclear if it will really fix the problem or just not make it worse.

lsf37:
At least it is now showing the failure here, so it did definitely fix something when I tested it.

corlewis (approved):

Issue #382, opened 2025-03-01
run `cmake --install` at end of compile


lsf37:
We're currently not testing the new `install` step of the build process. Would be easy to add to the kernel compile test and cheap in terms of runtime.

See also discussion on https://github.com/seL4/seL4/pull/1338

PR #383, opened 2025-03-02
standalone-kernel: test install target

lsf37:
Test cmake/ninja install target after compilation.

See also #382 and https://github.com/seL4/seL4/pull/1338

### DIFF
--- a/standalone-kernel/compile_kernel.sh
+++ b/standalone-kernel/compile_kernel.sh
@@ -97,17 +97,26 @@ do_compile_kernel()
         exit 1
     fi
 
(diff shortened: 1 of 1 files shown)


Indanz (approved):
It's better than nothing, but it doesn't test whether the install directory actually contains what it should.

lsf37:
> It's better than nothing, but it doesn't test whether the install directory actually contains what it should.

That part should happen in the kernel repo (ideally as a part of the install step. but could be a separate target if we want), because the scripts here wouldn't know which files those should be. This step here just makes sure `install` is called so that the hook is there.

PR #384, opened 2025-03-05
platforms: enable SMP tests on RPI4

lsf37:
After support in seL4/seL4#1420 and seL4/seL4_tools#222


Ivan-Velickovic:
`no_hw_test` is set so you can merge this now since compilation of SMP works fine. I can look into putting a RPi4B back into machine queue next week so we can do runtime tests as well.

Ivan-Velickovic (approved):

PR #388, opened 2025-03-11
platforms.yml: remove cheshire support

Ivan-Velickovic:
My bad, Cheshire doesn't have a timer so sel4bench can't build which means that we can't have it in platforms.yml

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -30,13 +30,6 @@ platforms:
     march: rv64imac
     no_hw_test: true
 
-  CHESHIRE:
-    arch: riscv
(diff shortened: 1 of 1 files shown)


lsf37 (approved):

kent-mcleod:
Its not impossible for devices to have no platform accessible timer. Is there a way for an exclusion for sel4bench support? There are also sel4bench apps that don't need a timer, so sel4bench should be able to be built without timer support too (with some changes).

lsf37:
If we can make sel4bench work for platforms without timer, that would be great. I just don't want to start getting half-finished platform ports for platforms that do have a timer.

PR #393, opened 2025-03-16
github: option access token for private repos

lsf37:
Input option for providing an access token for read-access, so that the checks can also work on private repositories.


lsf37:
This is mainly used in the l4v repo so far.

corlewis (approved):

PR #400, opened 2025-03-28
sel4bench: report IRQ invoke for context switch

midnightveil:
As @gernotheiser mentioned privately, the typical case should be for a /different/ address space, rather than the same address space; the first is uncommon and misrepresentative of the actual performance of seL4.


lsf37 (approved):
No issues with changing it from my side. Certainly seems to make more sense.

Just noting that this page has been online for about a decade and has been reviewed by Gernot multiple times. Code review really isn't worth much for spotting small things.

PR #402, opened 2025-04-09
github: pin gitlint in pr workflow to ubuntu 22.04

lsf37:
Running on Ubuntu 24.04 sometimes (not always) produces installation failure in the python setup step. Until that is sorted out, run the workflow only on Ubuntu 22.04.


corlewis (approved):
A nice example of this failure in the checks for this PR.

PR #403, opened 2025-04-09
fix gitlint & style lints

midnightveil:
Note: CI is expected to fail because it forces the GHA to use the version of the `master` branch. Which is horribly confusing.

### DIFF
--- a/scripts/install-python.sh
+++ b/scripts/install-python.sh
@@ -7,6 +7,8 @@
 # Installs python 3.10; assumes GitHub's ubuntu-latest
 
 echo "Installing python 3.10"
+# for python 3.10 for newer ubuntu distributions
+sudo add-apt-repository ppa:deadsnakes/ppa
 sudo apt-get install -qq python3.10 > /dev/null
 sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
 sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 1
--- a/scripts/install-python.sh
+++ b/scripts/install-python.sh
@@ -11,4 +11,9 @@ sudo apt-get install -qq python3.10 > /dev/null
 sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
 sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 1
 python3 --version
-pip3 install -q --upgrade pip setuptools==70.0.0 launchpadlib wheel
+
+# setuptools and wheel versions must be matched to working releases
+# we need an old version of setuptools (see PR seL4/ci-actions#381)
+# and newer versions of wheel removes the bdist_wheel implementation
+# that setuptools used to use, but no longer does in newer versions.
+pip3 install -q --upgrade pip setuptools==70.0.0 wheel==0.45.1 launchpadlib
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -42,7 +42,7 @@ jobs:
 
   style:
     name: Style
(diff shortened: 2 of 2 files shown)


lsf37:
Thanks, that does look promising. Let me push a test commit to that makes the action use the PR branch

Ivan-Velickovic (approved):

scripts/install-python.sh
```diff
@@ -7,8 +7,14 @@
...
 echo "Installing python 3.10"
+sudo add-apt-repository -q ppa:deadsnakes/ppa
 sudo apt-get install -qq python3.10 > /dev/null
 sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
 sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 1
 python3 --version
-pip3 install -q --upgrade pip setuptools==70.0.0 launchpadlib wheel
+
+# setuptools and wheel versions must be matched to working releases
+# we need an old version of setuptools (#381) and newer versions of
```
Ivan-Velickovic, line 17: the `#381` is a bit ambiguous

scripts/install-python.sh
```diff
@@ -7,8 +7,14 @@
 # Installs python 3.10; assumes GitHub's ubuntu-latest
 
 echo "Installing python 3.10"
+sudo add-apt-repository -q ppa:deadsnakes/ppa
```
lsf37, line 10: Apparently `add-apt-repository` does not have a `-q` option (see error message in https://github.com/seL4/ci-actions/actions/runs/14347901457/job/40221047571?pr=404).

```suggestion
# For python 3.10 on distributions that ship more recent pythons
sudo add-apt-repository ppa:deadsnakes/ppa
```

scripts/install-python.sh
lsf37: It's referring to https://github.com/seL4/ci-actions/pull/381 which is what I would expect from an unqualified issue number, so I think this one is fine.

Ivan-Velickovic: Okay no worries then.

lsf37:
> Thanks, that does look promising. Let me push a test commit to that makes the action use the PR branch

It mostly works -- see the run in https://github.com/seL4/ci-actions/actions/runs/14347901457/job/40221047571 which funnily is pointing out two gitlint fix-ups: both commit titles are 2 characters too long :-)

 It's still getting the default python, because of the `-q` in `add-apt-repository -q`, but the pinning is correct.

scripts/install-python.sh
```diff
@@ -7,8 +7,15 @@
...
+# for python 3.10 for newer ubuntu distributions
+sudo add-apt-repository -q ppa:deadsnakes/ppa
 sudo apt-get install -qq python3.10 > /dev/null
 sudo update-alternatives --install /usr/bin/python python /usr/bin/python3.10 1
 sudo update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.10 1
 python3 --version
-pip3 install -q --upgrade pip setuptools==70.0.0 launchpadlib wheel
+
+# setuptools and wheel versions must be matched to working releases
+# we need an old version of setuptools (see PR github:ci-actions#381)
```
lsf37, line 18: There is a defined GitHub shorthand for referring to repos:

```suggestion
# we need an old version of setuptools (see seL4/ci-actions#381)
```

scripts/install-python.sh
lsf37: The `-q` is still there

lsf37:
Cool. I'll add a commit for unpinning the style check, and then merge. Thanks for figuring this one out.

PR #406, opened 2025-04-28
sel4bench-web: generate json instead of html

lsf37:
Since the site is running on Jekyll it makes more sense to generate the page there directly and just provide the data as json input.

After the redesign, the new site is going to expect json input of for produce here.

### DIFF
--- a/sel4bench/build.py
+++ b/sel4bench/build.py
@@ -15,8 +15,6 @@ from builds import SKIP, SUCCESS, REPEAT, FAILURE
 from pprint import pprint
 from typing import List, Optional
 
-from datetime import datetime
-
 import json
 import os
 import sys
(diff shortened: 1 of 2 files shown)


Indanz (approved):
Looks good to me. Did you compare the JSON to the old html and verified the content is the same and in the correct place?

lsf37:
The Jekyll side is actually not implemented yet :), I ran out of time yesterday. But yes, making sure the output is unchanged is the plan.

lsf37:
Jekyll side now done and renders identical to before modulo the redundant clock column in the compilation details, which I have removed in a second commit here.

Indanz:
I meant to just check whether the numbers are the same and in the expected place, but an en-to-end check with rendered html is much better of course.

PR #408, opened 2025-05-16
platforms: enable hardware test for imx93

lsf37:
The board `imx93a` in now available in the machine queue.

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -171,7 +171,7 @@ platforms:
     aarch_hyp: [64]
     platform: imx93
     march: armv8a
-    no_hw_test: true
+    req: imx93a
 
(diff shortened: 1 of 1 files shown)


Indanz (approved):

seL4-platforms/platforms.yml
```diff
@@ -171,7 +171,7 @@ platforms:
     aarch_hyp: [64]
     platform: imx93
     march: armv8a
-    no_hw_test: true
+    req: imx93a
```
Indanz, line 174: Apologies, I should have noticed the missing `-DNUM_NODES=2` earlier, but I guess this is the place to add it.

lsf37: Right, yes, that should go in here, I think.

PR #410, opened 2025-05-26
platforms: fix typo

lsf37:
Fix the typo that is currently breaking the build everywhere (`:` instead of `=` in yaml, of course).


Indanz (approved):
Oops.

lsf37:
> Oops.

Yeah, I should know better than to do this kind of stuff while travelling..

Indanz:
And I should have checked the change better...

PR #415, opened 2025-06-30
builds: remove boot failure detection

lsf37:
The current code did not work (never got the full log to check boot failure on), and the mq/console scripts are already retrying on boot failure -- more retrying is unlikely to help and will just slow down test failure.

The hifive failures that started all this have been removed in the meantime.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -19,7 +19,6 @@ from typing import Optional, List, Tuple, Union
 from junitparser import JUnitXml
 
 import copy
-import time
 import os
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


Indanz (approved):
It does look weird, but correct.

I do have the feeling that the ssh wrapper check is at the wrong level. I'd expect it either lower or higher up. Now it seems easy to forget to add it everywhere it would make sense.

lsf37:
> It does look weird, but correct.
> 
> I do have the feeling that the ssh wrapper check is at the wrong level. I'd expect it either lower or higher up. Now it seems easy to forget to add it everywhere it would make sense.

The check only really makes sense for some of the `mq.sh` commands and the python entry points for that are all in this file. 

There are other ssh invocations in the tests, but they tend to be less intermittent. So for now I'm reasonably happy with the location of it.

PR #417, opened 2025-07-10
platforms.yml: add HiFive P550

Ivan-Velickovic:


seL4-platforms/platforms.yml
```diff
@@ -90,6 +90,14 @@ platforms:
     march: rv64imac
     no_hw_test: true
 
+  HIFIVE_P550:
+    arch: riscv
+    modes: [64]
+    smp: [64]
+    platform: hifive-p550
+    march: rv64imacfd
```
lsf37, line 98: Missed that one before: we need to call it `rv64imac` even if that's wrong, because it won't be picked up by the build matrix otherwise.

Really need to overhaul that field, it doesn't really have anything to do with the `march`

PR #420, opened 2025-07-11
platforms.yml: disable HiFive P550 SMP runs for now

Ivan-Velickovic:
When doing the initial port of the P550 to seL4, I saw failures when booting SMP but they were fixed after some improvements were made to the locking code by Indan. However, I was testing only in GCC and it looks like Clang has failures that must still be investigated.

See [1] for details.

[1]: https://github.com/seL4/seL4/actions/runs/16181893559/job/45764557154

### DIFF
--- a/seL4-platforms/platforms.yml
+++ b/seL4-platforms/platforms.yml
@@ -92,7 +92,6 @@ platforms:
   HIFIVE_P550:
     arch: riscv
     modes: [64]
-    smp: [64]
     platform: hifive-p550
     march: rv64imac
     req: [p550a]


lsf37 (approved):

Indanz:
I think it [this](https://github.com/seL4/seL4_tools/blob/b305af4dd306630ecfd2776d41d54b6d9964055c/elfloader-tool/src/arch-riscv/boot.c#L218) weird `i++` that's wrong.

Main entry hart_id:0, so it counts zero based. I think the loop should have been with i = either 0 or 1, depending on API:

```
    for (i = 1; i <= CONFIG_MAX_NUM_NODES && hsm_exists; i++) {
        if (i != hart_id) { // is this hart ID or core ID?
            sbi_hart_start(i, secondary_harts, i); // or i + 1 somewhere?
        }
    }
```
I suspect there is perhaps an off-by-one between core ID and hart and those two got confused somewhere?

PR #429, opened 2025-12-19
sel4bench: use early processing for signal benchmark results

JDuchniewicz:
Followup to #428 Depends on https://github.com/seL4/sel4bench/pull/68

Adds the result of early processing signal benchmark to be reported on the performance webpage.


lsf37 (approved):
We should observe a few runs and compare to the current numbers, but otherwise this is good to merge.

PR #450, opened 2026-04-07
sel4test-hw: increase verification skip scope

### LINKED ISSUE #380: Consider merging verification test configs with release configs because they're almost identical
I wonder whether the cost of running Verification separately from Release is worth any benefit

There's almost no distinction between them anymore.

Arm has: KernelArmHypEnableVCPUCP14SaveAndRestore was the only option I could find that would be on for release but off for verification

x86 has: KernelVTX, KernelX86_64VTX64BitGuests and KernelIOMMU as options that can't be on for verification but could be on otherwise.

RISCV has no difference.

So essentially the "release" option could be merged with the "verification" option in non-x86 platforms

And we should check and see why KernelArmHypEnableVCPUCP14SaveAndRestore can't be enabled with verification

Would save a bunch of time running tests

lsf37:
After corresponding PRs in seL4 and seL4test, increase the scope of skipping verification builds (and checking for config equivalences with release instead).

Needs to wait for https://github.com/seL4/sel4test/pull/154 to be merged and deployed before we can merge here.

### DIFF
--- a/sel4test-hw/build.py
+++ b/sel4test-hw/build.py
@@ -65,7 +65,8 @@ def hw_run(manifest_dir: str, build: Build):
 def verification_equals_release(build: Build) -> bool:
     """Return whether in this build release and verification settings are equivalent."""
 
-    return build.get_platform().arch == 'riscv'
+    plat = build.get_platform()
+    return plat.arch == 'riscv' or plat.arch == 'arm'
 
 
 def build_filter(build: Build) -> bool:


Indanz (approved):
For consistency, might want to retrieve `plat = build.get_platform()` once and use plat.arch twice, like you do elsewhere.

lsf37:
> For consistency, might want to retrieve `plat = build.get_platform()` once and use plat.arch twice, like you do elsewhere.

Good point, have done that now. The CI run still has not finished, but I think I can merge anyway, because re-runs will not see the new action code yet, and even if it does, this run will be the first commit combination where this actually works. So it should be safe (famous last words).

PR #452, opened 2026-04-10
preprocess: provide a base_branch option

lsf37:
Provide the option for checking against a specific base branch instead of the revision in the verification manifest.

Useful for pull-request checking when the base branch has moved on over the verification manifest.

This is the first half of implementing https://github.com/seL4/seL4/issues/1299. When this is merged, we still need to update the corresponding workflow file in the seL4 repo.

### DIFF
--- a/preprocess/steps.sh
+++ b/preprocess/steps.sh
@@ -19,7 +19,21 @@ checkout-manifest.sh
 
 REPOS="$(pwd)"
 SEL4_REPO="${REPOS}/seL4"
-MANIFEST_REV=$(git -C ${SEL4_REPO} rev-parse HEAD)
+
+REMOTE=$(git -C ${SEL4_REPO} remote)
+if [ "$(echo "${REMOTE}" | wc -l)" -ne 1 ]; then
+  echo "::error::Expected exactly one remote, got: ${REMOTE}"
+  exit 1
+fi
+
+if [ -n "${INPUT_BASE_REF}" ]
+then
+  # Use the tip of the specified base branch instead of the manifest hash
+  git -C ${SEL4_REPO} fetch -q --depth 1 "${REMOTE}" "${INPUT_BASE_REF}"
+  git -C ${SEL4_REPO} checkout -q FETCH_HEAD
+fi
+
+BASE_REV=$(git -C ${SEL4_REPO} rev-parse HEAD)
 
 cd "${SEL4_REPO}"
 export BRANCH_NAME="github-ci-work/pr-branch"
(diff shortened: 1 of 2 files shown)


lsf37:
The portable shell check failure is expected, this script does need bash and nullglob (not touched in this PR).

preprocess/steps.sh
```diff
@@ -19,7 +19,15 @@ checkout-manifest.sh
 
 REPOS="$(pwd)"
 SEL4_REPO="${REPOS}/seL4"
-MANIFEST_REV=$(git -C ${SEL4_REPO} rev-parse HEAD)
+
+if [ -n "${INPUT_BASE_BRANCH}" ]
+then
+  # Use the tip of the specified base branch instead of the manifest hash
+  git -C ${SEL4_REPO} fetch -q --depth 1 origin "${INPUT_BASE_BRANCH}"
```
Indanz: Is `origin` always correct here?

I would either expect an `INPUT_BASE_REV` which is always on `origin`, or remote + branch or full path or something like that (which is what SEL4_REPO seems to be).

Indanz (approved):
I suppose it does the right thing, but calling it a branch while it's a ref seems very confusing. Same applies to https://github.com/seL4/seL4/pull/1643.

preprocess/steps.sh
lsf37: Hm, you're right, in other places we do this we specifically add an `origin` remote to make sure. Actually, this is pretty much guaranteed to be wrong. We are always checking out the verification manifest in this one, and the remote that is set up there is `verification`, not `origin`, but we shouldn't really rely on that name here either. Need to think about that a bit.

lsf37: Ok, I've changed it to read the name of the remote out of `git remote`. We do know at this point that there will be only one remote set up, so even if we change the name in the manifest, we'll get the right one.

Have also change the input name to `base_ref` for consistency.

preprocess/steps.sh
```diff
@@ -19,7 +19,21 @@ checkout-manifest.sh
...
+REMOTE=$(git -C ${SEL4_REPO} remote)
+if [ "$(echo "${REMOTE}" | wc -l)" -ne 1 ]; then
+  echo "::error::Expected exactly one remote, got: ${REMOTE}"
+  exit 1
+fi
+
+if [ -n "${INPUT_BASE_REF}" ]
+then
+  # Use the tip of the specified base branch instead of the manifest hash
+  git -C ${SEL4_REPO} fetch -q --depth 1 "${REMOTE}" "${INPUT_BASE_REF}"
```
Indanz, line 32: I guess if it's a revision and there is only one remote, you probably don't need to specify which remote to use. That's mostly important when using branches.

PR #453, opened 2026-04-20
trivial: perform error checking

midnightveil:
Check that SSH keys as environment variables are set, and use `set -e` in a few scripts to check return codes.

Non-functional change.

Splitting out from https://github.com/au-ts/seL4-ci-actions/pull/1.

### DIFF
--- a/aws-proofs/steps.sh
+++ b/aws-proofs/steps.sh
@@ -46,6 +46,11 @@ if [ -z ${IP} ]; then exit 1; fi
 echo "Waiting for sshd to come up"
 until nc -w5 -z ${IP} 22; do echo "."; sleep 3; done
 
+if [ -z "${AWS_SSH}" ]; then
+  echo "No 'AWS_SSH' key provided" >&2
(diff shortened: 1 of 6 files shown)


Indanz (approved):
Assuming it doesn't break anything, it looks fine to me.

It there a clear rule when to use `set -e` and when not? Because now it seems quite ad-hoc. If only a few commands are expected to fail, perhaps mark those with e.g. `|| true` instead.

midnightveil:
> It there a clear rule when to use set -e and when not? Because now it seems quite ad-hoc. If only a few commands are expected to fail, perhaps mark those with e.g. || true instead.

I don't know. It was already ad-hoc; some scripts set it already. In my opinion we should always use `set -e` (and `-u`), but I'm aware that it (especially `-u`) will break some scripts.

lsf37 (approved):
I'm happy with that, and in particular with `set -e` as a default policy. I just haven't been very consistent with it.

PR #454, opened 2026-04-20
trivial: typos + missing input xml declaration

midnightveil:
Split out from https://github.com/au-ts/seL4-ci-actions/pull/1.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -544,7 +544,7 @@ def run_build_script(manifest_dir: str,
         result = SUCCESS
         output = None
         for line in script:
-            result, ouput = run_cmd(line, run, output)
+            result, output = run_cmd(line, run, output)
             if result != SUCCESS:
                 break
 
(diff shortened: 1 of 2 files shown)


seL4-platforms/builds.py
```diff
@@ -544,7 +544,7 @@ def run_build_script(manifest_dir: str,
         result = SUCCESS
         output = None
         for line in script:
-            result, ouput = run_cmd(line, run, output)
+            result, output = run_cmd(line, run, output)
```
Indanz, line 547: This doesn't look trivial, it changes actual behaviour, as the output of previous lines isn't fed to `run_cmd`.

How on Earth could the old code work?

seL4-platforms/builds.py
```diff
@@ -556,7 +556,7 @@ def run_build_script(manifest_dir: str,
         # run final script tasks even in case of failure, but not for SKIP
         if result != SKIP:
             for line in final_script:
-                r, output = run_cmd(line, run, ouput)
+                r, output = run_cmd(line, run, output)
```
Indanz, line 559: Same here.

seL4-platforms/builds.py
midnightveil: It looks like tbe only time that I can see that ever passes something other than a list of strings is in builds.py which does 'lambda r: SKIP'. I guess nothing uses this previous output feature...?

lsf37: I was planning to try to detect things like boot failures based on the output of the previous command, but I don't think I ever got around to actually implementing anything useful there. It might be better to remove the entire feature, but I keep thinking that it might be useful down the road, so I haven't done that yet :-)

In any case it is definitely better to use it correctly, even if it currently doesn't do anything.

lsf37 (approved):

PR #455, opened 2026-04-20
manifest deploy/checkout: support non-seL4 repos

midnightveil:
Also support branches other than 'master', to use names such as 'main'.

You can specify e.g. `manifest_repo: au-ts/microkit-manifest`.

Split out from https://github.com/au-ts/seL4-ci-actions/pull/1. Not strictly necessary to be merged in here, but useful for developing on a private fork.

### DIFF
--- a/manifest-deploy/steps.sh
+++ b/manifest-deploy/steps.sh
@@ -30,12 +30,19 @@ sudo apt-get install -qq doxygen
 echo "::endgroup::"
 
 echo "::group::Repo checkout"
-export MANIFEST_URL="ssh://git@github.com/seL4/${INPUT_MANIFEST_REPO}.git"
-export REPO_MANIFEST=master.xml
+if echo "$INPUT_MANIFEST_REPO" | grep -q "/" 2>/dev/null; then
+  export MANIFEST_URL="ssh://git@github.com/${INPUT_MANIFEST_REPO}.git"
+else
+  export MANIFEST_URL="ssh://git@github.com/seL4/${INPUT_MANIFEST_REPO}.git"
+fi
+export REPO_BRANCH="${INPUT_MANIFEST_BRANCH}"
+export REPO_MANIFEST="${INPUT_MANIFEST}"
 export REPO_DEPTH=0
 checkout-manifest.sh
 repo-util hashes
 echo "::endgroup::"
+# releaseit always writes to 'default.xml'
+unset REPO_MANIFEST
 
 echo "::group::Deploy"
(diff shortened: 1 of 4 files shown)


manifest-deploy/steps.sh
```diff
@@ -30,8 +30,13 @@ sudo apt-get install -qq doxygen
 echo "::endgroup::"
 
 echo "::group::Repo checkout"
-export MANIFEST_URL="ssh://git@github.com/seL4/${INPUT_MANIFEST_REPO}.git"
-export REPO_MANIFEST=master.xml
+if [[ "$INPUT_MANIFEST_REPO" =~ "/" ]]; then
```
Indanz: The double bracket syntax is Bash specific. Not a problem here, as it does #!/bin/bash, but just mentioning it.

manifest-deploy/steps.sh
```diff
@@ -30,8 +30,13 @@ sudo apt-get install -qq doxygen
...
+  export MANIFEST_URL="ssh://git@github.com/${INPUT_MANIFEST_REPO}.git"
+else
+  export MANIFEST_URL="ssh://git@github.com/seL4/${INPUT_MANIFEST_REPO}.git"
+fi
+export REPO_MANIFEST="$INPUT_MANIFEST"
```
Indanz: Deploy has no manifest input?

midnightveil: This was here before. The answer, of course, is that in Github actions if you pass an input not declared in action.yml it still gets passed through anyway as an environment variable, just with a warning.

I can check that it is used an add it to action.yml, though.

midnightveil: Now that I look at it it probably doesn't support it, I'll remove.

The source of `releaseit` is private though. The default would have to be `default.xml` not `master.xml`, so yep.

midnightveil: ... And I've thought this through more, no. `REPO_MANIFEST` is for `checkout-manifest.sh` not for releaseit. I've added that back and made the documentation clearer on that.

Indanz: > This was here before.

I meant INPUT_MANIFEST. It wasn't in manifest-deploy/action.yml, but I missed you added "manifest" to the inputs (even though I checked for that, weird).

midnightveil: I added that when I forced push today, so you weren't wrong.

lsf37 (approved):
Thanks for cleaning that up a bit.

PR #457, opened 2026-04-20
Avoid storing binaries in $HOME for local runners

midnightveil:
If we use a local runner to build, the builds are not sandboxed. This is mostly fine for this use, but let's avoid installing files in global folders. Instead namespace it under $GITHUB_WORKSPACE which is unique to a build.

Part of https://github.com/au-ts/seL4-ci-actions/pull/1, but not necessary to merge.
Just a nicety.

### DIFF
--- a/manifest-deploy/steps.sh
+++ b/manifest-deploy/steps.sh
@@ -8,12 +8,14 @@
 set -e
 
 echo "::group::Setting up"
-echo "Installing 'repo'"
-mkdir -p ~/bin
-curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
-chmod a+x ~/bin/repo
 
-PATH=~/bin:"${GITHUB_WORKSPACE}/seL4_release":$PATH
+echo "Installing 'repo'"
+BINDIR="${RUNNER_TEMP}/bin"
+mkdir -p "${BINDIR}"
+curl https://storage.googleapis.com/git-repo-downloads/repo > "${BINDIR}/repo"
+chmod a+x "${BINDIR}/repo"
+
+PATH="${BINDIR}":"${GITHUB_WORKSPACE}/seL4_release":$PATH
 
 if [ -z "${GH_SSH}" ]; then
   echo "No 'GH_SSH' key provided" >&2
--- a/l4v-deploy/steps.sh
+++ b/l4v-deploy/steps.sh
@@ -8,11 +8,13 @@
 set -e
 
 echo "::group::Setting up"
-mkdir -p ~/bin
-curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
-chmod a+x ~/bin/repo
 
-PATH=~/bin:"${SCRIPTS}/../l4v-deploy":$PATH
+BINDIR="${RUNNER_TEMP}/bin"
+mkdir -p "${BINDIR}"
+curl https://storage.googleapis.com/git-repo-downloads/repo > "${BINDIR}/repo"
+chmod a+x "${BINDIR}/repo"
+
+PATH="${BINDIR}":"${SCRIPTS}/../l4v-deploy":$PATH
 
 pip3 install --user lxml
 
--- a/repo-checkout/steps.sh
+++ b/repo-checkout/steps.sh
@@ -9,10 +9,12 @@ set -e
 
 echo "::group::Setting up"
 
-mkdir -p ~/bin
-curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
-chmod a+x ~/bin/repo
-PATH=~/bin:$PATH
+BINDIR="${RUNNER_TEMP}/bin"
+mkdir -p "${BINDIR}"
+curl https://storage.googleapis.com/git-repo-downloads/repo > "${BINDIR}/repo"
+chmod a+x "${BINDIR}/repo"
+
+PATH="${BINDIR}":$PATH
 
 pip3 install -U PyGithub
 


Indanz (commented):
At this point probably better to introduce a variable for `BIN="${GITHUB_WORKSPACE}/bin"` and use that instead of all the duplication.

lsf37:
Hm, I think this is problematic -- we can't assume that the workspace directory is empty and that `bin` is available in there. It might for instance already have a checkout at the point the action is called. Maybe not for these specific actions since they mostly check out their own stuff, but I don't think it's an assumption we should bake in.

I did so far assume sandboxed ephemeral runners, and it is true that all this breaks down for self hosted ones.

Should we use something in `/tmp` instead? (e.g. with `mktmp`).

Indanz:
> Hm, I think this is problematic -- we can't assume that the workspace directory is empty and that `bin` is available in there.

But that's what the current code already assumes for ~/bin, so in that sense this is strictly an improvement over existing behaviour.

> Should we use something in `/tmp` instead? (e.g. with `mktmp`).

I don't know how it works, but surely there is already a local directory where all outputs can be put, and I assume that's the purpose of GITHUB_WORKSPACE.

lsf37:
> > Hm, I think this is problematic -- we can't assume that the workspace directory is empty and that `bin` is available in there.
> 
> But that's what the current code already assumes for ~/bin, so in that sense this is strictly an improvement over existing behaviour.

No, on the GitHub runners, the home directory is separate from the workspace directory. The home directory has nothing in it and is ephemeral (but only on the GitHub runners, the same is not true on self-hosted ones, which is why Julia was hitting this problem).

> > Should we use something in `/tmp` instead? (e.g. with `mktmp`).
> 
> I don't know how it works, but surely there is already a local directory where all outputs can be put, and I assume that's the purpose of GITHUB_WORKSPACE.

GITHUB_WORKSPACE is where repositories are checked out and the action can do its work. It's up to the workflow that calls the action to organise the contents of this directory. Sometimes repos are checked out in a subdirectory, sometimes (when you just use the standard actions/checkout), the repo contents get dumped directly into GITHUB_WORKSPACE and there is no safe place in there to put other files unless you know the repo layout. The home directory is safe for that on GitHub runners, that's why I used it for that, but if we want to support general runners we need to be more careful.

midnightveil:
It looks like there is a RUNNER_TEMP directory https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context.

There's also a RUNNER_TOOL_CACHE: https://github.com/actions/setup-python/blob/c8813ba1bc76ebf779b911ad8ffccbf2e449cb48/docs/advanced-usage.md?plain=1#L549, though that's probably intended for tools that stay the same...

Indanz:
> GITHUB_WORKSPACE is where repositories are checked out and the action can do its work. It's up to the workflow that calls the action to organise the contents of this directory. Sometimes repos are checked out in a subdirectory, sometimes (when you just use the standard actions/checkout), the repo contents get dumped directly into GITHUB_WORKSPACE and there is no safe place in there to put other files unless you know the repo layout.

Yes, but those workflows use parts of ci-actions. All you need to do is reserve a directory for ci-actions output and that can be in GITHUB_WORKSPACE. Then all workflows using ci-actions know not to use that directory. Maybe "bin" is not a good name, but "ci-actions" should be fine, and binaries could be put into GITHUB_WORKSPACE/ci-actions/bin.

Doing workflow related stuff outside the GITHUB_WORKSPACE seems like a hack to avoid conflict between arbitrary workflows using ci-action and not the proper solution to this namespace problem. Doing it in /tmp is slightly better than using ~/, but still messy I think.

lsf37:
> > GITHUB_WORKSPACE is where repositories are checked out and the action can do its work. It's up to the workflow that calls the action to organise the contents of this directory. Sometimes repos are checked out in a subdirectory, sometimes (when you just use the standard actions/checkout), the repo contents get dumped directly into GITHUB_WORKSPACE and there is no safe place in there to put other files unless you know the repo layout.
> 
> Yes, but those workflows use parts of ci-actions. All you need to do is reserve a directory for ci-actions output and that can be in GITHUB_WORKSPACE. Then all workflows using ci-actions know not to use that directory. Maybe "bin" is not a good name, but "ci-actions" should be fine, and binaries could be put into GITHUB_WORKSPACE/ci-actions/bin.

That's not realistic. When you have a workflow file that needs to use the standard actions/checkout for some reason, you want to use the defaults and not mess around with having to remember which directory something needs to be in. You then also need to communicate to the action where the repository is etc. In those cases you really do want the repo to be the workspace root, and you also don't want to start introducing rules that repos can't have a certain subdirectory or something like that. 

> Doing workflow related stuff outside the GITHUB_WORKSPACE seems like a hack to avoid conflict between arbitrary workflows using ci-action and not the proper solution to this namespace problem. 

We do definitely want to support mixing arbitrary actions with actions from ci-actions. Almost all of the nontrivial workflow files need to do that. GITHUB_WORKSPACE is supposed to be the work space, not arbitrary storage. 

> Doing it in /tmp is slightly better than using ~/, but still messy I think.

I do agree that `~/` is messy. It just happened to be easy.

lsf37:
> It looks like there is a RUNNER_TEMP directory https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#runner-context.

`RUNNER_TEMP` sounds promising, I'd be happy with that one.
 
> There's also a RUNNER_TOOL_CACHE: https://github.com/actions/setup-python/blob/c8813ba1bc76ebf779b911ad8ffccbf2e449cb48/docs/advanced-usage.md?plain=1#L549, though that's probably intended for tools that stay the same...

It also seems to be intended for python things specifically, so probably not that one.

lsf37:
We could also have a look what actions like setup-python or setup-ghc etc do, but I'd also be happy with just RUNNER_TEMP, it sounds pretty much exactly like what we want.

midnightveil:
This has been changed to use RUNNER_TEMP now.

Indanz (approved):
Considering repo tries to download and update itself anyway, what is the advantage of downloading it manually?

midnightveil:
We need to install it somehow. It's either this or with apt (which would only work for Ubuntu)

lsf37:
> Considering repo tries to download and update itself anyway, what is the advantage of downloading it manually?

The thing that we download here is just the top-level invocation/installer. What it downloads afterwards itself is the basically the git repo of of the repo tool.

manifest-deploy/steps.sh
```diff
@@ -9,11 +9,11 @@ set -e
 
 echo "::group::Setting up"
 echo "Installing 'repo'"
-mkdir -p ~/bin
-curl https://storage.googleapis.com/git-repo-downloads/repo > ~/bin/repo
-chmod a+x ~/bin/repo
+mkdir -p "${GITHUB_WORKSPACE}/bin"
+curl https://storage.googleapis.com/git-repo-downloads/repo > "${GITHUB_WORKSPACE}/bin/repo"
+chmod a+x "${GITHUB_WORKSPACE}/bin/repo"
```
lsf37: This one is still on GITHUB_WORKSPACE instead of RUNNER_TEMP. Or was there something different about manifest-deploy that I forgot?

midnightveil: Yep, I just missed it when updating.

lsf37 (approved):

PR #458, opened 2026-04-20
Dockerfile: set up a virtual environment

midnightveil:
The Docker deploy actions were failing due to break-system-packages python 'feature' that exists in newer debians.

Note: this fixes the docker builds. I have yet to fix the other uses of `pip3 install`, which I also need to make operate in a virtual environment now.

### DIFF
--- a/camkes-test/Dockerfile
+++ b/camkes-test/Dockerfile
@@ -12,10 +12,14 @@ FROM trustworthysystems/camkes-cakeml-rust:latest
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-       libffi-dev \
+       libffi-dev python3-venv \
     && apt-get clean autoclean \
     && apt-get autoremove --yes \
     && rm -rf /var/lib/{apt,dpkg,cache,log}/
+
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
 RUN pip3 install "junitparser==3.*" PyGithub
 
 ARG ACTION
--- a/camkes-vm/Dockerfile
+++ b/camkes-vm/Dockerfile
@@ -12,10 +12,14 @@ FROM trustworthysystems/camkes:latest
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
-       libffi-dev \
+       libffi-dev python3-venv \
     && apt-get clean autoclean \
     && apt-get autoremove --yes \
     && rm -rf /var/lib/{apt,dpkg,cache,log}/
+
+ENV VIRTUAL_ENV=/opt/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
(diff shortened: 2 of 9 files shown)


lsf37 (approved):

lsf37:
Thanks. I'll merge this and see what else we need to update.

midnightveil:
@lsf37 I believe you're going to need https://github.com/seL4/ci-actions/pull/460. This has definitely broken any of the scripts that used `--user`.

lsf37:
> @lsf37 I believe you're going to need #460. This has definitely broken any of the scripts that used `--user`.

Most of those don't run on docker images, though, just on the bare runner. I.e. they should not be affected by any of the changes on the docker side.

midnightveil:
Ok. I wasn't personally clear on how many of the actions ran exclusively in docker vs in a github actions Ubuntu or in both, it wasn't clear to me.

(Also I have no idea how to properly test all users of any of these).

lsf37:
> Ok. I wasn't personally clear on how many of the actions ran exclusively in docker vs in a github actions Ubuntu or in both, it wasn't clear to me.

You can tell by the action.yml file: when it says `using: 'docker'` then it's a docker-based action, and when it says something like `main: '../js/index.js'`, then it's a bare javascript action (which in this repo pretty much always just calls `steps.sh` directly)

> (Also I have no idea how to properly test all users of any of these).

That one is indeed not so easy. When I do that, I usually scan all .github/workflows/ directories in all seL4 repos. Many of the actions are used only once, but some are used often.

lsf37:
I think I made a mistake in predicting the outcomes of this one. While it makes sense to do the `venv` for the bare GitHub actions so that we don't interfere with the runner system packages (i.e. #460), if we do the same in docker, we are now missing all the packages that the base docker images are providing. In particular all of `sel4-deps` and `camkes-deps` and whatever else the base build containers do. This is why the simulation jobs in https://github.com/seL4/seL4/pull/1592 are failing.

We could re-install `sel4-deps` etc in the `venv`, but that just increases the image size and duplicates everything. The point of using a docker image is that we are controlling the system packages and there is nothing else that we can be interfering with. So I think the right approach is to revert this PR, do `--break-system-packages` instead in docker, but to use `venv` when we're outside docker and might be interfering with something else (i.e. #460 is still fine).

midnightveil:
> We could re-install sel4-deps etc in the venv, but that just increases the image size and duplicates everything. The point of using a docker image is that we are controlling the system packages and there is nothing else that we can be interfering with. So I think the right approach is to revert this PR, do --break-system-packages instead in docker, but to use venv when we're outside docker and might be interfering with something else (i.e. https://github.com/seL4/ci-actions/pull/460 is still fine).

Yep, fair.

The other thing we could do that might be nicer for the docker images: is there are `apt` packages for junitparser and for pygithub that we could install. That wouldn't solve `sel4-deps` but...

lsf37:
> The other thing we could do that might be nicer for the docker images: is there are `apt` packages for junitparser and for pygithub that we could install. That wouldn't solve `sel4-deps` but...

I'd usually prefer the python-managed versions, because it's easier to pin the exact version numbers when things break (e.g. we need junit 3 specifically at the moment).

PR #459, opened 2026-04-21
Support Microkit HW CI

midnightveil:
- Adds seL4-platforms knowledge about microkit
- Adds actions for use by Microkit CI

Should be merged before https://github.com/seL4/microkit/pull/475.

The 'test' here is rather simple for the moment: it simply runs the 'Hello, World!' example (which works only in debug) and checks the output matches. The goal is to eventually improve this, but it at least tests that the debug boot flow works; which is better than nothing.

- [x] do we want to set me/~~Ivan~~ (Ivan is leaving) up as codeowners for this folder?

### DIFF
--- a/microkit-hw-build/steps.sh
+++ b/microkit-hw-build/steps.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright 2021, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# generic steps to invoke hardware builds on the machine queue
+# expects a standard `build.py --hw` invocation to work in directory INPUT_ACTION_NAME
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+if [ -z "${VIRTUAL_ENV}" ]; then
+  python3 -m venv "${GITHUB_WORKSPACE}/venv"
+  . "${GITHUB_WORKSPACE}/venv/bin/activate"
+fi
+pip3 install "junitparser==3.*" sel4-deps
+
+echo "::endgroup::"
+
+# start builds
+python3 "${ACTION_DIR}/${INPUT_ACTION_NAME}/build.py"
--- a/microkit-hw-build/steps.sh
+++ b/microkit-hw-build/steps.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+#
+# Copyright 2026, UNSW
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# generic steps to invoke hardware builds on the machine queue
+# expects a standard `build.py --hw` invocation to work in directory INPUT_ACTION_NAME
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+. ${SCRIPTS}/setup-python-venv.sh
+pip3 install "junitparser==3.*" sel4-deps
+
+echo "::endgroup::"
+
+# start builds
+python3 "${ACTION_DIR}/${INPUT_ACTION_NAME}/build.py"
--- a/microkit-hw-matrix/steps.sh
+++ b/microkit-hw-matrix/steps.sh
@@ -1,0 +1,36 @@
+#!/bin/bash
+#
+# Copyright 2021, Proofcraft Pty Ltd
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# generic steps to invoke hardware builds on the machine queue
+# expects a standard `build.py --hw` invocation to work in directory INPUT_ACTION_NAME
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+if [ -z "${VIRTUAL_ENV}" ]; then
+  python3 -m venv "${GITHUB_WORKSPACE}/venv"
+  . "${GITHUB_WORKSPACE}/venv/bin/activate"
+fi
+pip3 install "junitparser==3.*" sel4-deps
+
+echo "::endgroup::"
+
+cd microkit
+  python3 build_sdk.py --sel4=../seL4 --matrix=../build_sdk_matrix.json
+
+  export TEST_CASES=$(cat ../build_sdk_matrix.json)
+
+  echo "test_cases=${TEST_CASES}" >> "${GITHUB_OUTPUT}"
+cd -
+
+# exports the gh_output
+python3 "${ACTION_DIR}/${INPUT_ACTION_NAME}/build.py" --matrix
--- a/microkit-hw-matrix/steps.sh
+++ b/microkit-hw-matrix/steps.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+#
+# Copyright 2026, UNSW
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+# generic steps to invoke hardware builds on the machine queue
+# expects a standard `build.py --hw` invocation to work in directory INPUT_ACTION_NAME
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+. ${SCRIPTS}/setup-python-venv.sh
+pip3 install "junitparser==3.*" sel4-deps
+
+echo "::endgroup::"
+
+cd microkit
+  python3 build_sdk.py --sel4=../seL4 --matrix=../build_sdk_matrix.json
+
+  export TEST_CASES=$(cat ../build_sdk_matrix.json)
+
+  echo "test_cases=${TEST_CASES}" >> "${GITHUB_OUTPUT}"
+cd -
+
+# exports the gh_output
+python3 "${ACTION_DIR}/${INPUT_ACTION_NAME}/build.py" --matrix
--- a/microkit-hw-run/README.md
+++ b/microkit-hw-run/README.md
@@ -1,0 +1,9 @@
+<!--
+     Copyright 2021, Proofcraft Pty Ltd
+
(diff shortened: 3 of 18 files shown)


microkit-hw-matrix/steps.sh
```diff
@@ -0,0 +1,32 @@
...
+# expects a standard `build.py --hw` invocation to work in directory INPUT_ACTION_NAME
+
+set -e
+
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+pip3 install --user "junitparser==3.*" sel4-deps
```
midnightveil: TODO: for me to fix, install in virtual-env

midnightveil: fixed

lsf37:
> * [ ]  do we want to set me/Ivan up as codeowners for this folder?

Yes, that would definitely make sense.

seL4-platforms/platforms.yml
```diff
@@ -29,13 +29,25 @@ platforms:
...
+
+  SERENGETI:
+    arch: riscv
+    modes: [64]
+    platform: cheshire
+    req: [serengeti1, serengeti2]
+    march: rv64imac
+    # microkit hw build/test only
+    no_hw_build: true
+    no_hw_test: true
```
lsf37, line 51: I don't remember, was there a reason we're not testing serengti with plain sel4test as well?

seL4-platforms/platforms.yml
```diff
@@ -189,6 +211,29 @@ platforms:
...
+
+  # microkit only
+  IMX8MP_IOTGATE:
+    arch: arm
+    modes: [64]
+    smp: [64]
+    platform: imx8mp-iotgate
+    req: [iotgate1, iotgate3, iotgate4, iotgate5]
+    march: armv8a
+    # microkit only, no seL4
```
lsf37, line 233: Should we at least build a sel4test image? I'm uncomfortable with having boards that are not officially supported on seL4 but then supported in microkit. We should try to eliminate those cases, ideally by having full support on the seL4 side.

seL4-platforms/platforms.yml
```diff
@@ -277,11 +324,32 @@ platforms:
...
     # ElfloaderImage: "binary"
 
+  KRIA_K26:
+    arch: arm
+    modes: [64]
+    aarch_hyp: [64]
+    platform: zynqmp
+    march: armv8a
+    # not in seL4 tooling
+    no_hw_build: true
```
lsf37, line 340: I don't understand how it is possible to have a microkit build for the board when there is no seL4 tooling for it. Are there things missing for sel4test?

microkit-hw-run/README.md
```diff
@@ -0,0 +1,9 @@
+<!--
+     Copyright 2021, Proofcraft Pty Ltd
+
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# seL4Test Hardware Test Runs
+
+todo
```
lsf37: Please fill in :-)

microkit-hw-run/build.py
```diff
@@ -0,0 +1,211 @@
+# Copyright 2021, Proofcraft Pty Ltd
+# Copyright 2026, UNSW
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+"""
+Parse builds.yml and run sel4test hardware builds and runs on each of the build definitions.
```
lsf37:
```suggestion
Parse builds.yml and run microkit hardware builds and runs on each of the build definitions.
```

lsf37: Actually there isn't a builds.yml for this one, so the comment should instead describe what we're doing in this case.

microkit-hw-run/README.md
lsf37: This would be a good place to describe how the actions differ from the other ones, in particular that the full set of builds is defined in a json file in the micrkokit repo, which is then passed in to the action (rather than a builds.yml file in the ci-actions repo).

This is fine to do, we should just document where the data is.

The other two directories should also get a short README, potentially with a link to this one.

microkit-hw-run/build.py
```diff
@@ -0,0 +1,211 @@
...
+
+    boards = sorted(set([(b.microkit_board, b.microkit_march) for b in builds]))
+    matrix = {
+        "include": [{"board": board, "march": march} for (board, march) in boards]
+    }
+
+    return "gh_matrix=" + json.dumps(matrix)
+
+
+# If called as main, run all builds from builds.yml
```
lsf37: needs updating

microkit-hw-matrix/steps.sh
```diff
@@ -0,0 +1,36 @@
...
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+if [ -z "${VIRTUAL_ENV}" ]; then
+  python3 -m venv "${GITHUB_WORKSPACE}/venv"
+  . "${GITHUB_WORKSPACE}/venv/bin/activate"
+fi
```
lsf37: when we do the `venv` refactor, we shouldn't forget this one

microkit-hw-build/steps.sh
```diff
@@ -0,0 +1,28 @@
...
+echo "::group::Setting up"
+export ACTION_DIR="${SCRIPTS}/.."
+export PYTHONPATH="${ACTION_DIR}/seL4-platforms"
+
+# python env
+sudo apt-get install -y --no-install-recommends libffi-dev
+if [ -z "${VIRTUAL_ENV}" ]; then
+  python3 -m venv "${GITHUB_WORKSPACE}/venv"
+  . "${GITHUB_WORKSPACE}/venv/bin/activate"
+fi
```
lsf37: (marker for `venv` refactor)

lsf37 (approved):
This looks very reasonable overall, only some missing README/comments.

I have questions about the sel4test/microkit support mismatch, but we can merge this and figure that part out later.

seL4-platforms/platforms.yml
Ivan-Velickovic: The Kria26 is just the ZCU102 with a different memory layout and default UART address, so everything in the kernel side is already supported, and a DTS overlay is used in Microkit to adjust those two things.

seL4-platforms/platforms.yml
Ivan-Velickovic: Similar to the Kria, this platform is the same as the `imx8mp-evk` which seL4 has explicit support for but with a slightly different memory layout so we have a custom DTS overlay in Microkit that gets applied.

seL4-platforms/platforms.yml
Ivan-Velickovic: Right now Serengeti is actually the exact same platform as Cheshire. I merged the Microkit support because I thought there might be changes to Serengeti that would make running Cheshire images on Serengeti incompatible but right now there aren't any. Perhaps we should not have Serengeti as an explicit platform in Microkit, it's a bit of a premature generalisation right now.

Ivan-Velickovic:
> I have questions about the sel4test/microkit support mismatch, but we can merge this and figure that part out later.

I made a comment for each one but the answer to all of them is basically that seL4 doesn't want to have explicit support for every single minor variation of a particular SoC (e.g slightly different memory layout or UART address). However, since seL4 still needs to know all this information at build time, and since we have to have a pre-built kernel it means we have to add explicit boards for each variation in Microkit for now (hence the Kria26 and imx8mp-iotgate support).

seL4-platforms/platforms.yml
midnightveil: (There would be a difference in the sense that for Serengeti it would be possible to setup the sel4_libs timer: as serengeti has a timer peripheral where Cheshire does not.)

seL4-platforms/platforms.yml
midnightveil: It would be possible to add these to the seL4 tests. Just at tbe cost of evem more things to build and test. But maybe that's fine and should be done, IDK.

seL4-platforms/platforms.yml
Ivan-Velickovic: That's true, I should clarify *from seL4's perspective* there's no difference to my knowledge (memory layout, UART, SoC support etc).

lsf37:
> > I have questions about the sel4test/microkit support mismatch, but we can merge this and figure that part out later.
> 
> I made a comment for each one but the answer to all of them is basically that seL4 doesn't want to have explicit support for every single minor variation of a particular SoC (e.g slightly different memory layout or UART address). However, since seL4 still needs to know all this information at build time, and since we have to have a pre-built kernel it means we have to add explicit boards for each variation in Microkit for now (hence the Kria26 and imx8mp-iotgate support).

Ok, thanks for explaining, that all makes sense. The main thing to decide then is if the trivial differences are worth checking in sel4test, which is mainly a matter of build time (and possibly board availability). We can think about that separately from this PR, though, i.e. no reason to hold up anything here.

microkit-hw-run/README.md
```diff
@@ -0,0 +1,63 @@
...
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+
+# Microkit Hardware Test Runs
+
+This action runs test examples for the [microkit] hardware test platforms.
+The tests can be restricted to specific `board`, `config`, and `march` configurations,
+which we use to create a GitHub actions runner matrix.
+
+The test expects pre-build images produced by the [microkit-hw-build] action,
```
lsf37:
```suggestion
The test expects pre-built images produced by the [microkit-hw-build] action,
```

microkit-hw-run/README.md
```diff
@@ -0,0 +1,63 @@
...
+and performs the necessary steps to run this image on the [machine queue] at
+UNSW.
+
+The Microkit HW builds & tests are different to many of the other actions in
+this repository; rather than the builds being driven by a 'build.yml' and
+'platforms.yml' file within this repository, it extracts a list of test cases
+/ platforms from the Microkit SDK build system directly. This is the `TEST_CASES`
+variable that expects a JSON file used in both [microkit-hw-build] and here.
+
+These action knows how to build an example from the Microkit SDK, which is
```
lsf37:
```suggestion
This action knows how to build an example from the Microkit SDK, which is
```

microkit-hw-run/README.md
```diff
@@ -0,0 +1,63 @@
...
+also notably pre-built outside of the ci-actions action scripts; instead these
+scripts simply built the examples for each platform to run.
```
lsf37: In this sentence I'm getting confused what "these scripts" refers to. The scripts in ci-actions or scripts in microkit? The scripts in ci-actions just invoke `make` inside the microkit repo, right?

lsf37 (approved):
Thanks for adding the docs. A few very minor queries, happy for you to address/answer them and then merge directly.

microkit-hw-run/README.md
midnightveil: These in ci-actions. Let me reword.

PR #460, opened 2026-04-21
python: always install in a virtualenv

midnightveil:
This stops break-system-packages warnings and works better in self-hosted runner environments which aren't sandboxed.

Friend to https://github.com/seL4/ci-actions/pull/458

### DIFF
--- a/aws-proofs/steps.sh
+++ b/aws-proofs/steps.sh
@@ -13,6 +13,7 @@ if [ "${GITHUB_EVENT_NAME}" = "pull_request_target" ] ||
    [ "${GITHUB_EVENT_NAME}" = "pull_request" ]
 then
   echo "::group::PR info"
+  . ${SCRIPTS}/setup-python-venv.sh
   pip3 install -U PyGithub
   export INPUT_EXTRA_PRS="$(get-prs)"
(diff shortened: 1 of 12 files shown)


aws-proofs/steps.sh
```diff
@@ -13,6 +13,10 @@ if [ "${GITHUB_EVENT_NAME}" = "pull_request_target" ] ||
    [ "${GITHUB_EVENT_NAME}" = "pull_request" ]
 then
   echo "::group::PR info"
+  if [ -z "${VIRTUAL_ENV}" ]; then
+    python3 -m venv "${GITHUB_WORKSPACE}/venv"
+    . "${GITHUB_WORKSPACE}/venv/bin/activate"
+  fi
```
lsf37: Two things:

-  we should factor this out into a short script that we can source everywhere (in scripts/). Need to make sure that directory is available, but it should be available in all non-docker actions
-  `GITHUB_WORKSPACE` is not a good place for `venv`, esp not for these actions, because some of them do assume an existing repository checkout under `GITHUB_WORKSPACE`. I think something in  `/tmp` with `mktemp` would work.

lsf37 (approved):

PR #461, opened 2026-04-22
platforms: handle SIGINT interruption by exiting

midnightveil:
This makes it easier to use this locally, as otherwise it keeps retrying new tests.

See also: https://github.com/seL4/machine_queue/pull/33

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -21,6 +21,7 @@ from junitparser import JUnitXml
 import copy
 import os
 import shutil
+import signal
 import subprocess
 import sys
 
@@ -394,6 +395,7 @@ FAILURE = 0
 SUCCESS = 1
 SKIP = 2
 REPEAT = 3
+INTERRUPTED = 4
 
 # string for detecting ssh failure
 ssh_failure_str = "Unable to ssh to tftp.keg.cse.unsw.edu.au"
@@ -420,7 +422,7 @@ def success_from_bool(success: bool) -> int:
         return FAILURE
 
 
-def run_cmd(cmd, run: Union[Run, Build], prev_output: Optional[str] = None) -> int:
+def run_cmd(cmd, run: Union[Run, Build], prev_output: Optional[str] = None) -> tuple[int, list[str]]:
     """If the command is a List[str], echo + run command with arguments, otherwise
     expect a function, and run that function on the supplied Run plus outputs from
     previous command."""
@@ -440,6 +442,15 @@ def run_cmd(cmd, run: Union[Run, Build], prev_output: Optional[str] = None) -> i
             sys.stdout.flush()
             sys.stderr.flush()
         ret = process.wait()
+
+        # Follow the Bash WCE protocol and propagate SIGINT upwards
+        # ... https://www.cons.org/cracauer/sigint.html
+        # A negative value -N indicates the process was terminated by the signal N.
+        if ret < 0 and (-ret) == int(signal.SIGINT):
+            return INTERRUPTED, lines
+        elif ret == 130:
+            # 130 is another return status that is commonly used to mean SIGINT
+            return INTERRUPTED, lines
 
         return success_from_bool(ret == 0), lines
     else:
@@ -552,6 +563,8 @@ def run_build_script(manifest_dir: str,
             printc(ANSI_RED, ">>> command failed, aborting.")
         elif result == SKIP:
             printc(ANSI_YELLOW, ">>> skipping this test.")
+        elif result == INTERRUPTED:
+            printc(ANSI_RED, ">>> command interrupted (SIGINT), exiting after cleanup")
 
         # run final script tasks even in case of failure, but not for SKIP
         if result != SKIP:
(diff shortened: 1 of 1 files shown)


seL4-platforms/builds.py
```diff
@@ -420,7 +422,7 @@ def success_from_bool(success: bool) -> int:
         return FAILURE
 
 
-def run_cmd(cmd, run: Union[Run, Build], prev_output: Optional[str] = None) -> int:
+def run_cmd(cmd, run: Union[Run, Build], prev_output: Optional[str] = None) -> tuple[int, list[str]]:
```
Indanz, line 425: This is just a bugfix unrelated to the other changes.

Indanz (approved):
Slightly confusing because you don't actually install a signal handler for SIGINT, but try to infer it from return codes of subprocesses, so can never be quite sure it was a SIGINT and not just a normal error exit. But that's the whole point of this, to avoid retrying interrupted runs.

midnightveil:
> Slightly confusing because you don't actually install a signal handler for SIGINT, but try to infer it from return codes of subprocesses, so can never be quite sure it was a SIGINT and not just a normal error exit. But that's the whole point of this, to avoid retrying interrupted runs.

Well, there is a signal handler in Python, used to generate a KeyboardInterrupt. I haven't quite figured out how this interacts with Popen exiting.

One thing I'll note is the 130 is somewhat bash-subprocess specific: it is the error code bash reports in `$?` when it dies to SIGINT (specifically 128+n). If there were other shells that our RemoteConnandOn in mq.sh ran in over ssh we might get a different number. (apparently ksh uses 256+n reported in `$?`, which sounds like it would break when you do "exit `$?`" but...). Other alternative way is that in mq.sh we check if Remote ssh commands return 130 and then SIGINT itself (like I already have to do for 255 because ssh sometimes just reports that instead with connection muxing) that would remove some hardcoding here, in place of somewhere else  But ... I doubt there's a perfect solution anyway.

Indanz:
This PR assumes that SIGINT signals reach the final child, terminate it, and that that process propagates up. An alternative implementation would be to catch SIGINT in the top-level script and to cleanup the children from there.

midnightveil:
> This PR assumes that SIGINT signals reach the final child, terminate it, and that that process propagates up. An alternative implementation would be to catch SIGINT in the top-level script and to cleanup the children from there.

Maybe I need to investigate it more, but I have never seen a KeyboardInterrupt raised in Python while doing this. Which doesn't make sense to me at all, but is what I observed.  There is some [weird behaviour with subprocess where it was 250ms sfter SIGINT before killing the child more permanently](https://github.com/python/cpython/blob/d9444844d7f322334c001c1a301c560d88c78e76/Lib/subprocess.py#L882C11-L882C12) but I would expect to see KeyboardInterrupt but I don't.

But yes we could also put the subprocess in a different process group and handle tbis manually.

midnightveil:
As I've discovered, we've already hit another case where our machine queue scripts would cause a 255 code that's not because of a SIGINT: https://github.com/seL4/machine_queue/pull/33/changes#r3128036287.

I'm going to rework this.

```
Booting all finished, dropped to user space
panicked at /home/julia/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rkyv-0.8.12/src/option.rs:38:17:
called `ArchivedOption::unwrap()` on a `None` value
(aborted)
Found thread has no fault handler while trying to handle:
unknown fault
in thread 0x887fc07400 "rootserver" at address 0x2219e8
With stack:
0x24dc80: 0x221bd8
0x24dc88: 0x24dcd8
0x24dc90: 0x221b5c
0x24dc98: 0x221b5c
0x24dca0: 0x221b30
0x24dca8: 0x24dcd8
0x24dcb0: 0x22d550
0x24dcb8: 0x24f2ec
0x24dcc0: 0x0
0x24dcc8: 0x2055c8
0x24dcd0: 0x67
0x24dcd8: 0x24dcc8
0x24dce0: 0x205600
0x24dce8: 0x1
0x24dcf0: 0x0
0x24dcf8: 0x221118

[[Timeout]]
None

console_run returned -1
Shutting down zcu102_2

Command terminated by signal 2
0.07user 0.07system 2:17.80elapsed 0%CPU (0avgtext+0avgdata 7440maxresident)k
0inputs+16outputs (0major+11686minor)pagefaults 0swaps
>>> command interrupted (SIGINT), exiting after cleanup
+++ mq.sh sem -signal zcu102_2 -k microkit-locally-local-local-0
-----------[ end test ZCU102_debug ]-----------
::endgroup::
Traceback (most recent call last):
  File "/home/julia/ts/microkit-mmu/ci-actions/scripts/../microkit-hw-run/build.py", line 206, in <module>
    sys.exit(run_builds(builds, hw_run))
             ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julia/ts/microkit-mmu/ci-actions/seL4-platforms/builds.py", line 882, in run_builds
    results[run_fun(manifest_dir, build)].append(build.name)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/julia/ts/microkit-mmu/ci-actions/scripts/../microkit-hw-run/build.py", line 110, in hw_run
    return run_build_script(
           ^^^^^^^^^^^^^^^^^
  File "/home/julia/ts/microkit-mmu/ci-actions/seL4-platforms/builds.py", line 628, in run_build_script
    raise KeyboardInterrupt("received SIGINT during execution")
KeyboardInterrupt: received SIGINT during execution
```

PR #466, opened 2026-05-05
get-extra-prs: allow branches+sha in Test with

### LINKED ISSUE #198: Make "Test with:" support branches also
When working on forks and feature-branches, having "`Test with:`" support branches also besides just PRs would be nice. That avoids having to create PRs for every affected repo during the development cycle, but still allows using the github CI. The short syntax could be `<org>/<repo>@<branch>`, the long syntax is `https://github.com/<org>/<repo>/tree/<branch>`.

lsf37:
Extend "Test with" syntax with branches and commit hashes. get-prs now returns either `org/repo#num` or `org/repo@ref`.

The user-level syntax is "Test with org/repo@(branch-name|sha1)" or a GitHub URL to the branch/commit.


lsf37:
The regexp side is tested, but I still need to try actual full pull requests.  

Edit: GitHub testing now done in https://github.com/seL4/seL4/pull/1649, appears to be working.

Indanz (approved):

scripts/fetch-extra-prs.sh
```diff
@@ -4,17 +4,36 @@
...
+# Expects INPUT_EXTRA_PRS to be a space separated list of:
+#   org/repo#num  - pull request (fetched from refs/pull/<num>/head)
+#   org/repo@ref  - branch name or commit SHA (fetched as <ref>)
 
 if [ -z "${INPUT_XML}" ] && [ -n "${INPUT_EXTRA_PRS}" ]
 then
-  echo "extra PRs: ${INPUT_EXTRA_PRS}"
+  echo "extra refs: ${INPUT_EXTRA_PRS}"
 
   for PR in ${INPUT_EXTRA_PRS}
```
corlewis: Extremely minor but to be consistent with other changes you've made, `PR` could be `ref`, `INPUT_EXTRA_PRS` could be `INPUT_EXTRA_REFS` and the name of this file could be `fetch-extra-refs.sh`.

corlewis (approved):

scripts/fetch-extra-prs.sh
lsf37: You're right. I was trying to avoid update noise, but it's better to have consistent naming afterwards.

PR #475, opened 2026-05-22
repo-checkout: fix manifest depth argument

Ivan-Velickovic:
Whoops, sorry...


lsf37 (approved):

lsf37:
Interesting. With Julias change to ignore the signed-off-by lines, we now get actual enforcement of needing to have a commit message body. Previously the check was seeing the sign-off as a body and didn't actually enforce anything.

I'll override here for now. We'll see how annoying that becomes, maybe we should have some repos where that is explicitly not enforced.

midnightveil:
I noticed that too, had to add commit messages when rebasing some PRs.

It's probably a good thing, the rule was being violated in spirit and the check just happened to not care...

PR #477, opened 2026-05-25
sel4test-hw-run: enable x86 SMP + HYP

Ivan-Velickovic:
I've just removed the SMP check in the condition, I don't know whether we should also try remove verification + SMP being skipped.

### DIFF
--- a/sel4test-hw/build.py
+++ b/sel4test-hw/build.py
@@ -105,8 +105,8 @@ def build_filter(build: Build) -> bool:
             return False
 
     if plat.arch == 'x86':
-        # Bamboo config says no VTX for SMP or verification
-        if build.is_hyp() and (build.is_smp() or build.is_verification()):
+        # Bamboo config says no VTX for verification
+        if build.is_hyp() and build.is_verification():
             return False
 
     if plat.arch == 'riscv':


lsf37:
> I've just removed the SMP check in the condition, I don't know whether we should also try remove verification + SMP being skipped.

Let's leave that on for now. Build times are already pretty high and it's not a combination that makes much sense at the moment.

lsf37 (approved):

Indanz:
> > I've just removed the SMP check in the condition, I don't know whether we should also try remove verification + SMP being skipped.
> 
> Let's leave that on for now. Build times are already pretty high and it's not a combination that makes much sense at the moment.

That are all reasons to add it to the filter though. The filter says when to skip something.

The comment says "Bamboo config says". What Bamboo config is that? And should that be updated to match this filter? Don't we want to deliberately deviate from the Bamboo filter here, by compiling something but not running tests?

lsf37:
> > > I've just removed the SMP check in the condition, I don't know whether we should also try remove verification + SMP being skipped.
> > 
> > Let's leave that on for now. Build times are already pretty high and it's not a combination that makes much sense at the moment.
> 
> That are all reasons to add it to the filter though. The filter says when to skip something.

Yes, I'm saying we should still leave out SMP + verification from the test for now.
 
> The comment says "Bamboo config says". What Bamboo config is that? And should that be updated to match this filter? Don't we want to deliberately deviate from the Bamboo filter here, by compiling something but not running tests?

Bamboo is the CI system that we used before GitHub. I didn't write that one, and I was initially just trying to replicate what was there (the old code still exists in the private repo https://github.com/seL4/configascode if you like Java). We should probably remove references to Bamboo going forward, because it's not really relevant any more.

We should probably also remove all that xml junit printing and parsing (and possibly the Bamboo config setting in sel4test). Initially I thought we might want to extract more fine-grained test results from GitHub and display them somehow in the GitHub UI (as we could in Bamboo), but I don't think that can work with GitHub actions.

PR #479, opened 2026-05-26
link-check: use GITHUB_WORKSPACE for checkout

lsf37:
Use the `GITHUB_WORKSPACE` directory for repo checkout and link check run so that the generated `.lycheecache` is available outside the container for the github/cache action to pick up and make persistent.

### DIFF
--- a/link-check/steps.sh
+++ b/link-check/steps.sh
@@ -7,8 +7,6 @@
 set -e
 
 echo "::group::Setting up"
-mkdir -p /repo
-cd /repo
 checkout.sh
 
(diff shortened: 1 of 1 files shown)


midnightveil (approved):
I think by default we usually start in GITHUB_WORKSPACE anyway but being explicit harms no one.

lsf37:
> I think by default we usually start in GITHUB_WORKSPACE anyway but being explicit harms no one.

Yes, true, we could leave it out completely. Will do that.

PR #481, opened 2026-05-27
link-check: mark workspace dir as safe

lsf37:
git complains about ownership of the mounted workspace directory.

I have a feeling this was why I did the `mkdir /repo` thing initially.

### DIFF
--- a/link-check/steps.sh
+++ b/link-check/steps.sh
@@ -7,6 +7,8 @@
 set -e
 
 echo "::group::Setting up"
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

midnightveil:
It looks like the official GitHub checkout action also does the same thing, ha: https://github.com/actions/checkout/blob/main/dist/index.js#L1455-L1465

How wonderful....

PR #482, opened 2026-05-28
sel4test-hw: skip x86 32-bit SMP + hyp

Ivan-Velickovic:
64-bit seems to be fine, but 32-bit does not even compile:
```
In file included from /github/workspace/kernel/include/arch/x86/arch/kernel/elf.h:10,
                   from /github/workspace/kernel/src/arch/x86/32/kernel/elf.c:7:
  /github/workspace/kernel/include/arch/x86/arch/32/mode/kernel/elf.h:41:28: error: unterminated argument list invoking macro "NODE_STATE"
     41 | } Elf32_Phdr_t, Elf_Phdr_t;
        |                            ^
```

This compile error also occurs for 15.0.0 seL4, so it is not a recent regression.

PR #483, opened 2026-05-29
sel4test-hw-run: re-enable x86 32-bit hyp + vtx

midnightveil:
The kernel compile error should have been fixed now.

Ref: https://github.com/seL4/ci-actions/pull/482 and https://github.com/seL4/seL4/pull/1678


lsf37 (approved):
Let' try again.

PR #484, opened 2026-06-03
hw-tests: increase default timeout to 20min

lsf37:
Apparently two of the boards are slow enough to boot that they might hit 15min.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -59,7 +59,7 @@ class Build:
         self.success = None
         self.error = None
         self.settings = {}
-        self.timeout = 900
+        self.timeout = 1200
         self.no_hw_test = False
         self.image_base_name = "sel4test-driver"
         [self.name] = entries.keys()


midnightveil (approved):
Which boards, btw?

lsf37:
odroid_xu4 and tqma. I'm not sure how real that is. 15 min is a very long time to take for a sel4test run, but I'm happy to try it out.

lsf37:
(I think it's more likely that the output is getting lost, not that we're hitting a real timeout)

midnightveil:
Is there an example of the test that timed out where this was the issue?

Indanz:
From what I understood, because of the USB hub resetting at random times, we really should have whatever time that takes to resolve itself as a minimum timeout everywhere. But 15m sounds like something else though. because of the TFTP boot, it's probably a temporary network problem, but that should affect all boards equally.

> (I think it's more likely that the output is getting lost, not that we're hitting a real timeout)

Then how does increasing the timeout help? Wouldn't lowering it and then retrying more often be a better fix?

Also, how come a build timeout setting affects boot timeout???

midnightveil:
Is this sort of timeout the kind of thing we were seeing for ODROID_XU4? Assuming that this was the reason why the timeout was increased that makes no sense - especially as per the line-by-line timestamps GitHub records it waited for nearly 19minutes after this test.

https://github.com/seL4/seL4/actions/runs/26870228591/job/79277373021

It looks like based on sel4test that SCHED0000 is known flakey... is the test broken somehow?

---

Then again, the failure for the TQMA here where the output suddenly stopped was in a completely different test - it took 30s to get to that test per the timestamps, before waiting 19min to [[Timeout]].

https://github.com/seL4/seL4/actions/runs/26870228591/job/79277372831

(I wonder if 15minutes is even too long for one run tbh, but anyway).

Indanz:
That's just examples of the usual random CI breakage related to either consoles dropping out (because of USB hub reset), or other console problems (ssh or network or whatever). Both are probably worse if we do more tests and have more boards, because it seems to be a load problem or something. This has been going on for as long as I remember and it makes CI plain unreliable to detect rare actual problems.

lsf37:
> Then how does increasing the timeout help? Wouldn't lowering it and then retrying more often be a better fix?

I've just made the same argument to Alex :-). We might do that, I just wanted to try out if it actually is what Alex was thinking or not.

> Also, how come a build timeout setting affects boot timeout???

It's the runtime timeout, it just happens to be passed down from the build class, because the individual runs all get the same timeout. I know, not the most obvious choice.

midnightveil:
Should we revert this?

lsf37:
Yes. I was trying to figure out if we should reduce it, but haven't gotten around to it yet. Reverting is definitely fine, though.

PR #485, opened 2026-06-04
builds: reduce machine queue ssh connections

lsf37:
Reduce number of ssh connections to the machine queue, because the back-end server is running out of ssh sessions.

The previous info/lock/run/release cycle can all be done within just the `run` command, which can internally reduce the number of ssh connections that are made.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -266,16 +266,13 @@ class Run:
 
         return [
             ['tar', 'xvzf', f"../{self.build.name}-images.tar.gz"],
-            mq_print_lock(machine),
-            mq_lock(machine),
             mq_run(build.success, machine, build.files,
                    completion_timeout=build.timeout,
-                   lock_held=True,
                    key=job_key(),
                    log=log,
                    error_str=build.error)
         ], [
-            mq_release(machine)
+            # lock release is done in run command
         ]
 
 


lsf37:
I was also looking at reducing the number of connections for the cleanup jobs, but it turns out that none of those have been running for quite a while, because they fail at the junit install step. This explains why there are sometimes stales mq locks..

midnightveil:
TBH I don't actually understand how this would solve anything: these run in serial one after each other, so it would make in total less SSH connections, but the number in parallel must still be the same, no?

(And connection muxing means that it shouldn't need to redo the whole SSH handshake, just spawn a new pty on the server).

I'm more inclined to believe that the issue is that when a machine queue service is running the tftp server has a load average of ~3 just from SSH itself, and so is dropping things.

midnightveil (approved):
Let's give this a go anyway...

lsf37:
You're right that this will not be as much reduction as I initially thought, but overall I agree that we should try it, it can't be a bad thing.

lsf37:
Test running here: https://github.com/seL4/seL4/actions/runs/26923013130/job/79427245128

Indanz:
Well, I've seen errors like:

```
Failures for run #1, ignoring a whole bunch of console down failures:
2026-05-29T00:21:59.0563931Z [[Timeout]]
2026-05-29T00:21:54.4975803Z [[Timeout]]
2026-05-29T00:14:29.2003251Z ethernet@ff3f0000 Waiting for PHY auto negotiation to complete..console: FileRead(): SSL error on fd 3
2026-05-29T00:21:34.4250823Z console: FileRead(): SSL error on fd 3
2026-05-29T00:21:54.8962008Z [[Timeout]]
2026-05-29T00:04:40.4103796Z console: FileRead(): SSL error on fd 3
2026-05-29T00:08:31.5646626Z console: FileRead(): SSL error on fd 3
2026-05-29T00:19:15.2164458Z console: FileRead(): SSL error on fd 3
2026-05-29T00:31:10.4712253Z console: FileRead(): SSL error on fd 3
```
So there is some SSH related fishiness going on. (But this seems to be a new thing, on top of the old consoles disappearing misery.)

midnightveil:
Yes, seems like a new thing. That message is from conserver, and Peter was confused because none of our conserver connections use SSL between the different conserver instances

PR #489, opened 2026-06-09
Add GitHub summaries for failing jobs

lsf37:
Show a job summary table for jobs that have failures (but leave out successful runs to reduce noise).

The test summary data is recorded in a global dict that each job step incrementally adds data to.

Currently reports jUnit summaries and "Incomplete" for other step failures.

We could extend this (hopefully) fairly easily to do the kind of log analysis @midnightveil has been running on the test logs and print a more informative summary message in the table for "Incomplete". Another idea would be to include time stamps so that timeouts are easier to analyse. But I'd like to merge something working first.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -12,14 +12,15 @@ them, `run_build_script` for a standard test driver frame, and
 `default_junit_results` for a standard place to leave a jUnit summary file.
 """
 
-from junitparser.junitparser import Failure, Error
 from platforms import ValidationException, Platform, platforms, load_yaml, mcs_unsupported
 
 from typing import Optional, List, Tuple, Union
 from junitparser import JUnitXml
+from dataclasses import dataclass
 
 import copy
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


lsf37:
https://github.com/seL4/seL4/actions/runs/27186660605 should have summary tables once the re-run there finishes.

I haven't figured out yet what prompts GitHub to fold action summaries instead of having them expanded (size, number, etc), but even with folded summaries it is much faster to get to the error logs than before.

seL4-platforms/builds.py
```diff
@@ -453,33 +454,89 @@ def printc(color: str, content: str):
...
+    causes: List[Tuple[str, Optional[str]]] = field(default_factory=list)
 
-    Returns True if there were no failures or errors, raises exception
+    def success(self) -> bool:
+        # causes may include timeout or infrastructure failures that are not failed test cases
+        return self.failed == 0 and len(self.causes) == 0
+
+
+# Global dict, mapping run name to TestSummary for later display
+_junit_summaries: dict[str, TestSummary] = {}
```
lsf37: I should probably rename that to just `_test_summaries`, because the junit part is becoming increasingly inaccurate if we are adding more causes like timeout etc.

midnightveil (approved):

Indanz:
This is great, thanks for implementing this!

It's slightly glitchy, but still a great improvement compared to select failed board -> download raw log -> search for "FAILED" workflow. Issues I encountered:

- For me the summary page kept switching between two URLs after initially loading it, making it impossible to scroll etc. That stopped after I edited the URL in the URL bar manually (without hitting enter). (Firefox 151.0.2). This only happens when navigating to it via cogwheel -> "View job summary", next to the download raw log option. Example: https://github.com/seL4/seL4/actions/runs/27186660605/attempts/1#summary-80264549073. It works fine without the #* part. If I open the debug console, it stops and I see the following messages:
```
Too many calls to Location or History APIs within a short timeframe. 2 history.ts:51:18
Uncaught DOMException: The operation is insecure.
    819435/nB/history[e]/< history.ts:51
    update turbo.es2017-esm.js:2302
    e history.ts:54
    p history.ts:94
    a history.ts:37
    f history.ts:82
    checkForHashReset job-summaries.ts:28
    819435 events.ts:239
    v turbo.es2017-esm.js:383
    notifyApplicationAfterClickingLinkToLocation turbo.es2017-esm.js:3129
    applicationAllowsFollowingLinkToLocation turbo.es2017-esm.js:3121
    willFollowLinkToLocation turbo.es2017-esm.js:3036
    clickBubbled turbo.es2017-esm.js:1147
    nR progressive.ts:186
history.ts:51:18
```
So this seems weird interaction with Github's manual history management or something. Chrome works fine.

- The text "Incomplete" shows as a link, but goes nowhere, so doesn't seem to be valid. Example for PC99_debug_SMP_hyp_gcc_32: https://github.com/seL4/seL4/actions/runs/27186660605/attempts/1#pc99_debug_smp_hyp_gcc_32-incomplete

- The last summary has no Incomplete link, and also no excerpt from the failing log. I don't know whether that's because it's the last one, or of the nature of the ZYNQMP106 failure. Example: https://github.com/seL4/seL4/actions/runs/27186660605/attempts/1#summary-80264549388

All in all minor issues that can be fixed later.

seL4-platforms/builds.py
```diff
@@ -453,33 +454,89 @@ def printc(color: str, content: str):
...
+
+    with open(file_path) as f:
+        text = f.read()
+    blocks = []
+    for m in TESTCASE_RE.finditer(text):
+        block = m.group(0)
+        if "<failure" in block or "<error" in block:
+            name = NAME_RE.search(block)
+            blocks.append((name.group(1) if name else "", block))
+    return blocks
```
Indanz, line 506: I guess this is the place to add more checks, like for "Timeout" or boot failures.

Indanz (approved):

Indanz:
(The zynq "failed" with:

```
2026-06-09T07:09:48.0712129Z All is well in the universe
2026-06-09T07:09:48.0712316Z 
2026-06-09T07:09:48.0712320Z 
2026-06-09T07:09:48.1728081Z 
2026-06-09T07:09:48.1728587Z console_run returned 0
2026-06-09T07:09:48.1731912Z Shutting down zcu106
2026-06-09T07:09:48.1765182Z 
2026-06-09T07:09:51.0293124Z 0.54user 0.97system 1:00.94elapsed 2%CPU (0avgtext+0avgdata 8260maxresident)k
2026-06-09T07:09:51.0293940Z 0inputs+80outputs (0major+190607minor)pagefaults 0swaps
2026-06-09T07:09:51.0295081Z [33m+++ python3 ../projects/seL4_libs/libsel4test/tools/extract_results.py -q results.xml parsed_results.xml[0m
2026-06-09T07:09:51.1080480Z Failed to strip leading and trailing garbage
2026-06-09T07:09:51.1183345Z [31;1m>>> command failed, aborting.[0m
2026-06-09T07:09:51.1183968Z [33m+++ mq.sh sem -signal zcu106 -k seL4/seL4-Deploy-Test-27186660605-hw-run-37[0m
2026-06-09T07:09:55.8423136Z [1m-----------[ end test ZYNQMP106_debug_SMP_hyp_MCS_clang_64 ]-----------[0m
2026-06-09T07:09:55.8424486Z ##[endgroup]
2026-06-09T07:09:55.8425090Z [31;1mZYNQMP106_debug_SMP_hyp_MCS_clang_64 FAILED[0m
```
So the failure was caused by "Failed to strip leading and trailing garbage".

Maybe worth just printing the last N lines before FAILED, with a way to get the right ones instead of the summary ones.)

lsf37:
Annoyingly in the rerun everything succeeded, so I don’t know if my link fix worked :-). Will have a closer look tomorrow.

midnightveil:
I can "break" a board tomorrow if it's helpful, btw.

Indanz:
> Annoyingly in the rerun everything succeeded, so I don’t know if my link fix worked :-). Will have a closer look tomorrow.

There is a "Rerun all jobs" button for the last run. Don't know if that picks up any changes though.

lsf37:
> > Annoyingly in the rerun everything succeeded, so I don’t know if my link fix worked :-). Will have a closer look tomorrow.
> 
> There is a "Rerun all jobs" button for the last run. Don't know if that picks up any changes though.

It does, and one of your re-runs show that my fix did not work. Checking the generated HTML, there is just no anchor ID at all generated for the headings, so I'll have to insert my own.

lsf37:
> * For me the summary page kept switching between two URLs after initially loading it, making it impossible to scroll etc. That stopped after I edited the URL in the URL bar manually (without hitting enter). (Firefox 151.0.2). 

This seems to be a GitHub issue, i.e. there is nothing much I can do to the HTML of the main page, I can just write to the markdown file that GitHub designates for that job. It is possible that it is glitchy especially when jobs are still running and not all summaries are there yet.

> * The last summary has no Incomplete link, and also no excerpt from the failing log. I don't know whether that's because it's the last one, or of the nature of the ZYNQMP106 failure. Example: https://github.com/seL4/seL4/actions/runs/27186660605/attempts/1#summary-80264549388

That one is interesting: the output is empty, because the step that fails is the `sanitise_junit` step, which has no output of its own. It is correct to fail, because the test output (from the step before) is actually complete garbage. It ends like a successful test, but it just starts somewhere in the middle. I'm not even sure the output is from the correct board (although it might be, maybe just the beginning of the output got lost). In any case, that means a bit more thinking is necessary on how to analyse output. Just output of the current step isn't good enough. Should probably accumulate the output of all steps somewhere.

But I'll do that one later. For now I just want to fix the link issue and will then merge.

Indanz:
> This seems to be a GitHub issue, i.e. there is nothing much I can do to the HTML of the main page, I can just write to the markdown file that GitHub designates for that job.

It really messes up the history too, so can't go back either.

> It is possible that it is glitchy especially when jobs are still running and not all summaries are there yet.

Maybe, but I don't think so. I think it triggers a bug in the Github history code. Easy to check by going the the link when the runs are done though. Edit: Still happens while all tests are done, so that's not it.

seL4-platforms/builds.py
lsf37: done

lsf37:
Links are now working.

PR #492, opened 2026-06-11
builds.py: fix markdown table syntax

lsf37:
Number of header columns did not line up with next row.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -974,7 +974,7 @@ def job_failure_summary() -> Optional[str]:
         "## Test failures",
         "",
         "| Config | | Causes | Failed | Succeeded | Skipped |",
-        "| --- | --- | --- | --- | --- | --- | --- |",
+        "| --- | --- | --- | --- | --- | --- |",
     ]
     for summary in failed + passed:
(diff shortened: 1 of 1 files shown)


seL4-platforms/builds.py
```diff
@@ -974,7 +974,7 @@ def cause(config: str, cause: str, detail: Optional[str]) -> str:
         "## Test failures",
         "",
         "| Config | | Causes | Failed | Succeeded | Skipped |",
```
midnightveil, line 976: Isnt' there an empty column here then?

lsf37: There is, but I still had one too many before (I removed the separate "Total" column at one point and am now printing succeeded/total in the Succeeded column, because that takes less horizontal space and makes the table fit on the screen more often)

midnightveil: Yes. More my point was "why do we have the empty 6th column"?

midnightveil (approved):

seL4-platforms/builds.py
lsf37: Ah, it's the overall pass/fail column, but I wanted it small and the header Pass/Fail would be larger than any entry, as would be "Success", and since it's fairly obvious what it is, I just went without header for that column.

PR #493, opened 2026-06-11
thylint: add find_names to list of diagnostics

corlewis:
A find_names command is almost always for interactive diagnostics and not intended to be checked in.


lsf37 (approved):
Thank you.

PR #494, opened 2026-06-12
sel4test-hw: add build subgroups to speed up armv8 builds 

lsf37:
- add subgroups to the "march" build groups so we can parallelise specific groups independently
- use these for armv8a in sel4test-hw
- generate build matrix in python, so the subgroups get picked up automatically and don't have to be repeated

This speeds up the overall sel4test-hw build from over 50min to about 13min, which is the slowest riscv build. armv8 builds now finish just before riscv at about 12:30min. 

After this is merged, we still need to update the deployment workflow file in the seL4 repo, but the hw-test and hw-build PR checks should already benefit automatically.


Indanz (approved):
Cool, we needed to wait forever for AArch64 builds to finish, this should speed it up indeed.

PR #496, opened 2026-06-16
sel4bench: add results extraction and storage

lsf37:
- extract the results we want to store from the raw json output and append to time series files in sel4bench-results repo
- refactor web results extraction to use the new help functions
- add a no-sync option to checkout-manifest and set up git committer info so that it can be used in the steps above
- minor GH actions version fix that was missed before

This needs to be merged together with https://github.com/seL4/sel4bench/pull/74


sel4bench-metrics/steps.sh
```diff
@@ -0,0 +1,61 @@
...
+cd -
+echo "::endgroup::"
+
+# extract results and update sel4bench-results working copy
+python3 "${ACTION_DIR}/${INPUT_ACTION_NAME}/build.py" --metrics
+
+cd sel4bench-results
+git add .
+git commit -m "CI: add new benchmark results"
+git push main origin
```
midnightveil: `git push origin main`, no?

sel4bench/build.py
```diff
@@ -287,6 +244,73 @@ def gen_json(runs: List[Run], yml, file_name: str):
...
+    for run in runs:
+        results_file = f"results/{run.name}.json"
+        if not os.path.exists(results_file):
+            # not all potential runs are in the CI run/build matrix
+            continue
+
+        with open(results_file) as f:
+            data = json.load(f)
+
+        entry = dict(header)
```
midnightveil: I think this should probably be just `copy.deepcopy`. It's easier to understand and will mean we won't shoot ourselves in the foot later if we have nested things in the header.

sel4bench-metrics/steps.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "::group::Reading manifest"
+mkdir manifest
+cd manifest
+export MANIFEST_URL="https://github.com/seL4/sel4bench-manifest"
+export REPO_MANIFEST="default.xml"
+export REPO_BRANCH="${INPUT_MANIFEST_SHA}"
+export REPO_NO_SYNC="1"
+${SCRIPTS}/checkout-manifest.sh
+
```
midnightveil, line 44: If we're not going to actually sync the manifest wouldn't it make more sense to just clone to git repo directly?

Indanz (approved):

sel4bench-metrics/steps.sh
lsf37: We do potentially want `repo` to resolve the manifest, which can be constructed through multiple xml files. It turns out that in this specific case in the way I'm only looking at default.xml that's probably not necessary because default.xml tends to be fully resolved, but I didn't want to rely on that being always true in the future.

sel4bench-metrics/steps.sh
lsf37: yes, very much. The only line in this thing I have not tested so far :-)

PR #501, opened 2026-06-26
sel4bench-metrics: remove checkout-manifest dance

lsf37:
Instead of using Google repo and checkout-manifest, download the `default.xml` file directly.

I previously wanted to guard against default.xml potentially having references to other xml files, but that is unlikely and the whole idea of using `repo init -b <sha>` does not work.

@midnightveil was completely right, and that whole `checkout-repo.sh` thing was way too complicated. It also fails silently in CI, so it looked like everything was fine, when in fact it had never worked.

### DIFF
--- a/sel4bench-metrics/steps.sh
+++ b/sel4bench-metrics/steps.sh
@@ -15,16 +15,10 @@
 # - result extraction from json in results/
 # - git commit/push to sel4bench-results repo
 
+set -e
+
 echo "::group::Setting up"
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

Indanz:
For what's it worth, I have git submodule setup at https://github.com/Indanz/seL4-hello-world which I don't really keep up-to-date as it's a bit of a pain to do that manually all the time. But it's an example of how things can look like without repo.

PR #503, opened 2026-06-27
sel4bench: more robust json result extraction

lsf37:
Only extract the first "JSON OUTPUT" block in stdout.

It is possible for there to be more than one when the mq scripts fail to detect that the machine has booted correctly, the tests completes, boot detection times out, and another full test completes. In that case, there is no test failure and both results are valid.

This is generally rare, but it does happen every once in a while, recently on skylake twice, and there is no reason not to just use the result. It'd be nicer to make sure the boot detection is solid everywhere, but I don't want to dive into those scripts.

### DIFF
--- a/sel4bench/build.py
+++ b/sel4bench/build.py
@@ -51,9 +51,11 @@ def hw_build(manifest_dir: str, build: Build):
 def extract_json(results: str, run: Run) -> int:
     """Process test logs to extract JSON results."""
 
+    # Only extract the first JSON OUTPUT block. In rare cases (mq failure to
+    # detect boot completion), there can be two, and both are valid results.
     res = subprocess.run(
         f"cat {results} | "
-        f"sed -e '/JSON OUTPUT/,/END JSON OUTPUT/!d' | "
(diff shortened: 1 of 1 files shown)


Indanz (commented):
Now that the HW test CI setup is finally fixed and stable, this shouldn't be needed any more. All those boot problems and these kind of detection problems were probably all caused by the UART output disappearing at random moments. So I propose to postpone the merging of this one to easier see if it still happens.

Indanz (approved):
So although CI has improved, there is still about one in ten chance of it failing because of USB hub link reset.

lsf37:
> Now that the HW test CI setup is finally fixed and stable, this shouldn't be needed any more. All those boot problems and these kind of detection problems were probably all caused by the UART output disappearing at random moments. 

Yes, that is very likely correct

> So I propose to postpone the merging of this one to easier see if it still happens.

I'd still do use this, because it's one where there is perfect recovery and it does safe us from a manual intervention when it's not necessary. It does mask those kinds of failure, although if they are still there, we will see them in other tests.

PR #504, opened 2026-06-30
record and set sel4bench iterations

lsf37:
- update sel4bench-metrics code to record results for multiple benchmark iterations (for collecting noise data)
- add an `iterations` argument to the action to override default iteration count
- set default iteration count to 5 for now

We might want to reduce the iteration count to 3 if 5 takes too long, but the higher it is the faster we get good noise data.

This PR needs to be merged just after https://github.com/seL4/sel4bench-results/pull/3


lsf37:
We could also set `iterations: 1` or `iterations: 3` for PR checks and keep the default 5 for the deployment runs which record the noise.

For now we don't need iterations on the PR checks, but once we start detecting regressions, we will probably want at least 3.

Indanz (approved):

PR #505, opened 2026-07-01
sel4bench: scale hw-run timeout by iterations

lsf37:
Scale the hardware run timeout by 0.75 * iterations.

The current single-iteration timeout contains board boot time + about 2 * estimated benchmark run time, so scaling by 0.75 should be more than enough.

(The slowest benchmark config `PC99_SMP_64_haswell3 ` seems to be failing on legitimate timeouts sometimes with 5 iterations.)

### DIFF
--- a/.github/workflows/sel4bench-hw.yml
+++ b/.github/workflows/sel4bench-hw.yml
@@ -129,6 +129,7 @@ jobs:
           platform: ${{ matrix.platform }}
           req: ${{ matrix.req }}
           index: ${{ strategy.job-index }}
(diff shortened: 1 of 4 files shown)


midnightveil (approved):
I think strictly speaking this should be an affine transformation, not a linear one, but this will probably work fine for now (and it's not like we can really compute the constant factors)

lsf37:
Indeed. 0.75 is probably too high, actually, but I'll give this a shot for now and we can tweak it as we go along.

PR #507, opened 2026-07-01
Fix stdout print

Indanz:
Mistake introduced by previous commit (PR #506). We want to print `line` as it is, not decoded at all. Change this back to how it was.


alzer89 (approved):
Fixed string to bytes error

PR #508, opened 2026-07-06
platforms: detect and summarise common failures 

lsf37:
Detect infrastructure failure causes and seL4 crash failures and report them in the GitHub summary. Try to distinguish board boot failures from test hangs.

Builds on @midnightveil's work in https://github.com/midnightveil/seL4_logs

To get there, first:
-  refactor scripts for analysis steps
-  use analysis for repeat on ssh failure

We can probably go further than this with automatically analysing failures, but this is a good first step.

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -604,6 +604,8 @@ infra_failure_markers = [
 # component is the cause name in the summary. Earlier markers are matched first.
 sel4_failure_markers = [
     ("seL4 failed assertion", "Kernel assertion"),
+    ("Debug halt syscall from user thread", "Debug halt"),
+    ("Kernel entry via Unknown syscall", "Kernel entry via unknown syscall"),
 ]
 
 # Generic mq.sh completion-timeout marker. When this is present we will attempt
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -29,7 +29,8 @@ import time
 # exported names:
 __all__ = [
     "Build", "load_builds", "run_builds", "run_build_script", "junit_results", "sanitise_junit",
-    "sim_script", "Step", "default_analysis", "Analysis"
+    "sim_script", "Step", "Analysis", "default_analysis", "ssh_repeat_analysis", "infra_analysis",
+    "seL4_analysis"
 ]
 
 # where to expect jUnit results by default
@@ -231,8 +232,8 @@ class Build:
         return self.get_platform().getISA(self.get_mode())
 
     # create a Run on the fly if we only want one Run per Build
-    def hw_run(self, log):
-        return Run(self).hw_run(log)
+    def hw_run(self, log, analyse=None):
+        return Run(self).hw_run(log, analyse)
 
 
 class Run:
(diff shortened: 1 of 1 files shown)


seL4-platforms/builds.py
```diff
@@ -562,6 +551,172 @@ def sim_script(success: str, failure=None, timeout=1200):
...
+    ("[[Boot timeout]]", "boot failure"),
+]
+
+# kernel failure markers that are not standard junit test failures. Second
+# component is the cause name in the summary. Earlier markers are matched first.
+sel4_failure_markers = [
+    ("seL4 failed assertion", "Kernel assertion"),
+    ("Debug halt syscall from user thread", "Debug halt"),
+    ("Kernel entry via Unknown syscall", "Kernel entry via unknown syscall"),
+]
```
midnightveil, line 615: Would the

```
halting...
```

print be a good idea to add here?

Or just, "Kernel entry via Unknown" as my serror debug print PR.
(Similarly, I don't think this would likely catch kernel data abort?)

seL4-platforms/builds.py
```diff
@@ -562,6 +551,172 @@ def sim_script(success: str, failure=None, timeout=1200):
...
+# Generic mq.sh completion-timeout marker. When this is present we will attempt
+# to further classify.
+timeout_marker = "[[Timeout]]"
+
+# Console output detecting elfloader or seL4 is running. If none of these are
+# present, a timeout is a board boot problem.
+sel4_running_markers = [
+    # elfloader output
+    "Jumping to kernel-image entry point",
+    "Enabling MMU and paging",
```
midnightveil, line 626: Since this is used for Microkit too:

```
LDR|INFO: loader for seL4 starting
...
LDR|INFO|CPU0: jumping to kernel
...
MON|INFO: Microkit Monitor started!
```

seL4-platforms/builds.py
lsf37: I'm not sure about `halting...`, because it is very short and might appear as legitimate test output. The `Debug halt` is what caused the `halting...` in the case we looked at. That one is long enough, I think.

Agreed to shorten to `Kernel entry via Unknown`, that's a good idea.

I'll have a look at what we print for data abort.

lsf37: I changed my mind, let's add `halting...` as well

lsf37: Ok, now done.

seL4-platforms/builds.py
lsf37: Done.

midnightveil (approved):

seL4-platforms/builds.py
Indanz: For Arm Elfloader the text got changed recently to:

"Enabling MMU and jumping to entry point..."

and

"Enabling hypervisor MMU and jumping to entry point..."

Indanz (approved):

seL4-platforms/builds.py
lsf37: I'll add these

PR #510, opened 2026-07-06
sel4bench: better json/yaml file error reporting

lsf37:
When reading json or yaml, report which file was being read when there are any errors. Useful to diagnose corrupted result files in CI.

### DIFF
--- a/sel4bench/build.py
+++ b/sel4bench/build.py
@@ -78,6 +78,15 @@ def hw_build(manifest_dir: str, build: Build):
 def extract_json(results: str, run: Run) -> int:
     """Process test logs to extract JSON results."""
 
+    # Make sure there is at least one JSON OUTPUT block. This can happen when
+    # mq.sh fails catastrophically at its job, sel4bench gets interrupted, something
+    # else boots and runs instead, produces "All is well", but no sel4bench output.
+    # This should really not be necessary, but here we are.
+    if subprocess.run(f"grep -q 'JSON OUTPUT' {results}", shell=True).returncode != 0:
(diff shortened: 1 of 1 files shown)


lsf37:
Specifically, the error in https://github.com/seL4/sel4bench/actions/runs/28782831182/job/85356218231 has now happened 3 times, and it's hard to figure out which job actually failed anything.

(The actual test case should also be failing in this case, but that's a separate PR)

lsf37:
Now I accidentally pushed that part to this branch, sorry. If that's fine with you, let's just keep them together in this PR.

midnightveil (approved):

sel4bench/build.py
```diff
@@ -60,6 +78,15 @@ def hw_build(manifest_dir: str, build: Build):
...
     """Process test logs to extract JSON results."""
 
+    # Make sure there is at least one JSON OUTPUT block. This can happen when
+    # mq.sh fails catastrophically at its job, sel4bench gets interrupted, something
+    # else boots and runs instead, produces "All is well", but no sel4bench output.
+    # This should really not be necessary, but here we are.
+    if subprocess.run(f"grep -q 'JSON OUTPUT' {results}", shell=True).returncode != 0:
+        printc(ANSI_RED, f"Run {run.name} produced no JSON OUTPUT results.")
+        sys.stdout.flush()
+        return REPEAT
```
Indanz, line 88: More a generic remark I suppose, but is there a chance actions will be repeated forever, or is there some built in limit to prevent this?

lsf37: There is a global max_tries = 3 setting.

PR #513, opened 2026-07-08
platforms/builds: fix exception printing

lsf37:
The "+ e" is a type error, "{e}" should print using `__str__`.


midnightveil (approved):

midnightveil:
Is it worth using the backtrace library print backtrace function? So we get the full python stack trace of the inner exception?

i.e. `traceback.print_exc()`?

lsf37:
I guess we could, yes. Currently we would have gotten `lxml.etree.XMLSyntaxError: invalid character in attribute value, line 190, column 50` I think, but a backtrace would probably be useful.

Will add.

Indanz (approved):

PR #514, opened 2026-07-08
platforms/builds: add traceback print

lsf37:
Print the exception traceback for junit parsing failures. (addendum to #513)

### DIFF
--- a/seL4-platforms/builds.py
+++ b/seL4-platforms/builds.py
@@ -25,6 +25,7 @@ import shutil
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


midnightveil (approved):

Indanz (commented):
I don't think a backtrace would help, as it's an input problem, not a script problem. Printing out the input file name is enough I think. Of course, it also doesn't hurt.

lsf37:
With the one we just had it would not have helped, but if it's something obscure maybe it does.

PR #517, opened 2026-07-16
"Test with" for plain git checkouts

lsf37:
- Generalise `fetch-extra-refs.sh` so that it can pull branches or PRs for plain git clones that are not part of a repo manifest checkout. Use an env variable `EXTRA_REFS_PATHS` for org/repo to path mapping in that case.

- Make sure `repo-util` has empty stdout when repo manifest is not present.

- Use the new `fetch-extra-refs.sh` abilities to make "Test with:" specs available for camkes-unit tests.

- Add GitHub token input for decreasing rate limits on PR description fetch.

This should make the camkes unit tests work that were failing for the recent ticks/us PR.

### DIFF
--- a/camkes-unit/steps.sh
+++ b/camkes-unit/steps.sh
@@ -13,12 +13,22 @@ git clone https://github.com/seL4/capdl
 mkdir camkes-tool
 cd camkes-tool
 checkout.sh
+cd - > /dev/null
+
+if [ "${GITHUB_EVENT_NAME}" = "pull_request_target" ] ||
+   [ "${GITHUB_EVENT_NAME}" = "pull_request" ]
+then
+  export INPUT_EXTRA_REFS="$(get-prs)"
+  export EXTRA_REFS_PATHS="seL4/capdl=capdl"
+  fetch-extra-refs.sh
+fi
 
 echo
 echo "Repo summary:"
-echo "camkes-tool: $(git rev-parse --short HEAD)"
-echo "      capdl: $(git -C ../capdl rev-parse --short HEAD)"
+echo "camkes-tool: $(git -C camkes-tool rev-parse --short HEAD)"
+echo "      capdl: $(git -C capdl rev-parse --short HEAD)"
 
+cd camkes-tool
 export PYTHONPATH=../capdl/python-capdl-tool
 echo "::endgroup::"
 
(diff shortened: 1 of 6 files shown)


camkes-unit/steps.sh
```diff
@@ -13,12 +13,22 @@ git clone https://github.com/seL4/capdl
...
+  export EXTRA_REFS_PATHS="seL4/capdl=capdl"
+  fetch-extra-refs.sh
+fi
 
 echo
 echo "Repo summary:"
-echo "camkes-tool: $(git rev-parse --short HEAD)"
-echo "      capdl: $(git -C ../capdl rev-parse --short HEAD)"
+echo "camkes-tool: $(git -C camkes-tool rev-parse --short HEAD)"
+echo "      capdl: $(git -C capdl rev-parse --short HEAD)"
```
Indanz, line 29: This is just an echo, not real commands, so not sure what it does. If it's executed after this script is run then it needs to be unchanged, as at that point we did change to the camkes-tool dir.

scripts/fetch-extra-refs.sh
```diff
@@ -45,7 +83,7 @@ then
       URL="https://${REPO_PATH}"
     fi
 
-    cd $(repo-util path ${REPO})
+    cd "${DIR}"
```
Indanz, line 86: What's the difference between `repo_dir` and `repo-util path`?

Indanz (approved):
As far as I can judge it, it seems okay.

camkes-unit/steps.sh
lsf37: It's supposed to print out the state of the checked-out repos after all extra branch fetching and other operations, so that we can see in the logs what code the test actually ran on.

The change here is just because I changed the directory handling further up (cwd used to be camkes-tool at this point, now it is the parent directory of both, because that's what the branch fetching scripts expect).

scripts/fetch-extra-refs.sh
lsf37: `repo_dir <name>` is the generalised version of `repo-util path <name>`, where we first check for that new manual env variable, then fall back to `repo-util`. I wasn't entirely sure whether that would all be better done in python inside `repo-util` or outside like now, but I decided that it's nicer to keep `repo-util` focussed on just repo manifests.

camkes-unit/steps.sh
Indanz: But it's just an echo of the commands that would show that info? Is the output of this script executed by the caller?
Edit: Never mind, I missed the fancy inlined execution of shell commands.

midnightveil (approved):

